PR #9, opened 2020-03-16
#8 pulling publish date for a course

### LINKED ISSUE #8: getting the published date for a course
Course audit logs API consists of course information about course publish date as an event. In case of multiple publish date we always pull the latest. According to [API definition](https://canvas.instructure.com/doc/api/course_audit_log.html#method.course_audit_api.for_course), the order the events are published in Desc. So we will always get the recent published date
```
{
      "id": "3444",
      "created_at": "2020-03-16T19:37:29Z",
      "event_type": "published",
      "event_source": "api",
      "event_data": {
        
      },
```

Published Date look in `published` event

pushyamig:
fixes #8

### DIFF
--- a/inventory.py
+++ b/inventory.py
@@ -102,12 +102,10 @@ def handle_request_if_failed(response):
         return True
 
 
-def get_course_publish_unpublish_date(df_row: str, publish_or_unpublish_date, next_page_url = None)->str:
+def get_course_publish_date(df_row: str, publish_date, next_page_url = None)->str:
     logger.debug(df_row)
-    if df_row['course_workflow_state'] == 'available':
-        event_type_to_check_for = 'published'
-    if df_row['course_workflow_state'] == 'unpublished':
-        event_type_to_check_for = 'claimed'
+    if df_row['course_workflow_state'] != 'available':
+        return 'null'
 
     course_id= df_row['course_id']
 
@@ -120,34 +118,34 @@ def get_course_publish_unpublish_date(df_row: str, publish_or_unpublish_date, ne
         response = make_audit_logs_api_request(url)
 
     except (Exception) as e:
-        logger.exception('getting sections has erroneous response ' + e.message)
-        return publish_or_unpublish_date
+        logger.exception('getting published date for a course has erroneous response ' + e.message)
+        return publish_date
 
     if not handle_request_if_failed(response):
-        return publish_or_unpublish_date
+        return publish_date
 
     audit_events = json.loads(response.text.encode('utf8'))
     if not audit_events:
-        return publish_or_unpublish_date
+        return publish_date
 
     events = audit_events['events']
 
     for event in events:
-        if event['event_type'] == event_type_to_check_for:
-            publish_or_unpublish_date = event['created_at']
+        if event['event_type'] == 'published':
+            publish_date = event['created_at']
             logger.info(f"Date for Workflow type :{df_row['course_workflow_state']} for course {course_id}")
             break
-    if publish_or_unpublish_date is None:
+    if publish_date is None:
         next_page_url = get_next_page_url(response)
         if next_page_url is not None:
-            get_course_publish_unpublish_date(course_id, publish_or_unpublish_date, next_page_url)
+            get_course_publish_date(course_id, publish_date, next_page_url)
 
-    logger.info(f"For course {course_id} publish_or_unpublish_date: {publish_or_unpublish_date}")
-    if publish_or_unpublish_date is None:
+    logger.info(f"For course {course_id} publish_date: {publish_date}")
+    if publish_date is None:
         logger.info(f"For course {course_id} for workflow type {df_row['course_workflow_state']} don't have any date")
         return 'null'
 
-    return publish_or_unpublish_date
+    return publish_date
 
 
 def gather_course_info_for_account(account_id: int, term_id: int) -> Sequence[int]:
@@ -175,7 +173,7 @@ def gather_course_info_for_account(account_id: int, term_id: int) -> Sequence[in
     course_df = pd.DataFrame(slim_course_dicts)
     course_df['course_warehouse_id'] = course_df['course_id'].map(lambda x: x + WAREHOUSE_INCREMENT)
     course_date=None
-    course_df['unpub_published_date'] = course_df.apply(lambda x: get_course_publish_unpublish_date(x, course_date), axis=1)
+    course_df['published_date'] = course_df.apply(lambda x: get_course_publish_date(x, course_date), axis=1)
     logger.debug(course_df.head())
     course_df.to_csv(os.path.join('data', 'course.csv'), index=False)
     logger.info('Course data was written to data/course.csv')
--- a/inventory.py
+++ b/inventory.py
@@ -7,8 +7,6 @@ from typing import Dict, Sequence, Union
 import pandas as pd
 import psycopg2
 from umich_api.api_utils import ApiUtil
-from timeit import default_timer as timer
-from canvasapis import GroupsForSections
 import time
 from queue import Queue
 from threading import Thread
@@ -32,8 +30,6 @@ SUBSCRIPTION_NAME = ENV['API_SUBSCRIPTION_NAME']
 API_SCOPE_PREFIX = ENV['API_SCOPE_PREFIX']
 MAX_REQ_ATTEMPTS = ENV['MAX_REQ_ATTEMPTS']
 PER_PAGE_COUNT = ENV['PER_PAGE_COUNT']
-CANVAS_TOKEN = ENV['CANVAS_TOKEN']
-CANVAS_URL = ENV['CANVAS_URL']
 NO_OF_THREADS = ENV['NO_OF_THREADS']
 
 UDW_CONN = psycopg2.connect(**ENV['UDW'])
@@ -200,49 +196,6 @@ def get_course_publish_date_short(course_id)->str:
     strftime = time.strftime("%H:%M:%S", time.gmtime(seconds))
     logger.info(f"For course {course_id} API call took: {strftime}")
     return response
-
-
-def get_course_publish_date_direct_canvas(course_id, api_int)->str:
-    logger.debug(f"Getting Published date API call for course {course_id}")
-    start_time = time.time()
-    publish_date = None
-
-    url = f"{API_SCOPE_PREFIX}/audit/course/courses/{course_id}?per_page={PER_PAGE_COUNT}"
-
-    try:
-        response = api_int.get_course_audit_logs(course_id)
-
-    except Exception as e:
-        logger.exception('getting published date for a course has erroneous response ' + e.message)
-        return None
-
-    if not handle_request_if_failed(response):
-        return None
-    try:
-        audit_events = json.loads(response.text.encode('utf8'))
-    except JSONDecodeError as e:
-        logger.error(f"Json response Decoding failed due {e.msg}")
-        return None
-
-    if not audit_events:
-        return None
-
-    events = audit_events['events']
-
-    for event in events:
-        if event['event_type'] == 'published':
-            publish_date = event['created_at']
-            logger.info(f"Published Date {publish_date} for course {course_id}")
-            break
-
-    if publish_date is None:
-        logger.info(f"For course {course_id} don't seems to  have a date")
-        return None
-    seconds = time.time() - start_time
-    strftime = time.strftime("%H:%M:%S", time.gmtime(seconds))
-    logger.info(f"For course {course_id} API call took: {strftime}")
-    return publish_date
-
 
 def gather_course_info_for_account(account_id: int, term_id: int) -> Sequence[int]:
     # logger.info("gather_course_info_for_account")
--- a/inventory.py
+++ b/inventory.py
@@ -7,6 +7,12 @@ from typing import Dict, Sequence, Union
 import pandas as pd
 import psycopg2
 from umich_api.api_utils import ApiUtil
+from timeit import default_timer as timer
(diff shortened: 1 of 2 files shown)


inventory.py
```diff
@@ -25,13 +25,20 @@
...
 UDW_CONN = psycopg2.connect(**ENV['UDW'])
 WAREHOUSE_INCREMENT = ENV['WAREHOUSE_INCREMENT']
 
 
 # Function(s)
 
+def make_audit_logs_api_request(url, params=None):
+    logger.info(f"make_audit_logs_api_request url: {url}")
+    response = API_UTIL.api_call(url, SUBSCRIPTION_NAME, payload=params)
+    return response
```
ssciolla: It would be nice if we could combine some of your code here, and maybe the `handle_request_if_failed` functions, with the existing `make_request_using_api_utils`. Not sure how feasible that is, but it seems like there is a lot of duplication now. If we need to change my other paging approach to be more in line with yours, that would be okay.

inventory.py
```diff
@@ -69,6 +76,79 @@ def slim_down_course_data(course_data: Sequence[Dict]) -> Sequence[Dict]:
...
+
+
+def get_course_publish_unpublish_date(df_row: str, publish_or_unpublish_date, next_page_url = None)->str:
+    logger.debug(df_row)
+    if df_row['course_workflow_state'] == 'available':
+        event_type_to_check_for = 'published'
+    if df_row['course_workflow_state'] == 'unpublished':
+        event_type_to_check_for = 'claimed'
+
+    course_id= df_row['course_id']
```
ssciolla:
```suggestion
    course_id = df_row['course_id']
```

inventory.py
```diff
@@ -69,6 +76,79 @@ def slim_down_course_data(course_data: Sequence[Dict]) -> Sequence[Dict]:
...
+        url = next_page_url
+    else:
+        url = f"{API_SCOPE_PREFIX}/audit/course/courses/{course_id}?per_page={PER_PAGE_COUNT}"
+
+    try:
+        response = make_audit_logs_api_request(url)
+
+    except (Exception) as e:
+        logger.exception('getting sections has erroneous response ' + e.message)
+        return publish_or_unpublish_date
```
ssciolla: What exception are you seeing here? Any chance it's the `JSONDecodeError` I'm catching in the `make_request_using_api_utils` function?

inventory.py
```diff
@@ -69,6 +76,79 @@ def slim_down_course_data(course_data: Sequence[Dict]) -> Sequence[Dict]:
...
+    audit_events = json.loads(response.text.encode('utf8'))
+    if not audit_events:
+        return publish_or_unpublish_date
+
+    events = audit_events['events']
+
+    for event in events:
+        if event['event_type'] == event_type_to_check_for:
+            publish_or_unpublish_date = event['created_at']
+            logger.info(f"Date for Workflow type :{df_row['course_workflow_state']} for course {course_id}")
```
ssciolla:
```suggestion
            logger.info(f"Date for Workflow type {df_row['course_workflow_state']} for course {course_id}")
```

inventory.py
```diff
@@ -69,6 +76,79 @@ def slim_down_course_data(course_data: Sequence[Dict]) -> Sequence[Dict]:
...
+            break
+    if publish_or_unpublish_date is None:
+        next_page_url = get_next_page_url(response)
+        if next_page_url is not None:
+            get_course_publish_unpublish_date(course_id, publish_or_unpublish_date, next_page_url)
+
+    logger.info(f"For course {course_id} publish_or_unpublish_date: {publish_or_unpublish_date}")
+    if publish_or_unpublish_date is None:
+        logger.info(f"For course {course_id} for workflow type {df_row['course_workflow_state']} don't have any date")
+        return 'null'
```
ssciolla: I think it might make more sense for this to be `None`, rather than a string.

inventory.py
```diff
@@ -94,6 +174,8 @@ def gather_course_info_for_account(account_id: int, term_id: int) -> Sequence[in
 
     course_df = pd.DataFrame(slim_course_dicts)
     course_df['course_warehouse_id'] = course_df['course_id'].map(lambda x: x + WAREHOUSE_INCREMENT)
+    course_date=None
```
ssciolla:
```suggestion
    course_date = None
```

inventory.py
```diff
@@ -25,13 +25,20 @@
 SUBSCRIPTION_NAME = ENV['API_SUBSCRIPTION_NAME']
 API_SCOPE_PREFIX = ENV['API_SCOPE_PREFIX']
 MAX_REQ_ATTEMPTS = ENV['MAX_REQ_ATTEMPTS']
+PER_PAGE_COUNT = ENV['PER_PAGE_COUNT']
```
ssciolla: Maybe we should also plug this into my request function?

inventory.py
```diff
@@ -69,6 +76,79 @@ def slim_down_course_data(course_data: Sequence[Dict]) -> Sequence[Dict]:
         slim_course_dicts.append(slim_course_dict)
     return slim_course_dicts
 
+def get_next_page_url(response):
```
ssciolla: Please add a newline.

inventory.py
```diff
@@ -69,6 +76,79 @@ def slim_down_course_data(course_data: Sequence[Dict]) -> Sequence[Dict]:
...
+
+    for event in events:
+        if event['event_type'] == event_type_to_check_for:
+            publish_or_unpublish_date = event['created_at']
+            logger.info(f"Date for Workflow type :{df_row['course_workflow_state']} for course {course_id}")
+            break
+    if publish_or_unpublish_date is None:
+        next_page_url = get_next_page_url(response)
+        if next_page_url is not None:
+            get_course_publish_unpublish_date(course_id, publish_or_unpublish_date, next_page_url)
```
ssciolla: How does this work here? The recursion idea is kinda cool, but I don't see you using the results from querying the next page?

inventory.py
```diff
@@ -69,6 +76,77 @@ def slim_down_course_data(course_data: Sequence[Dict]) -> Sequence[Dict]:
...
+    if response.status_code != 200:
+        return False
+    else:
+        return True
+
+
+def get_course_publish_date(df_row: str, publish_date, next_page_url = None)->str:
+    logger.debug(df_row)
+    if df_row['course_workflow_state'] != 'available':
+        return 'null'
```
zqian: return None here instead of 'null'

inventory.py
```diff
@@ -69,6 +76,77 @@ def slim_down_course_data(course_data: Sequence[Dict]) -> Sequence[Dict]:
...
+        return publish_date
+
+    events = audit_events['events']
+
+    for event in events:
+        if event['event_type'] == 'published':
+            publish_date = event['created_at']
+            logger.info(f"Date for Workflow type :{df_row['course_workflow_state']} for course {course_id}")
+            break
+    if publish_date is None:
```
zqian: so the logic is keep looking for the "published" event, until we find it; otherwise, we can stop there.

I understand the people can iterator through publish vs unpublish status, so this approach will get the earliest publish date. That is fine with me.

In this case, there is then no need to have the "publish_datet" param inside the get_course_publish_date() function, since the recursive function will stop once the date is found.

inventory.py
pushyamig: Your code is returning date object after fetching data when you have to page through using link header then you need link headers for that. Here is the code [sample](https://github.com/tl-its-umich-edu/api-utils-python/blob/master/tests/test.py#L80) for that.

Yeah, redundant code but approach for pulling course un/ publish data is different from getting published date. It can be consolidated for the future. 

@zqian what is the long term scope of this project? will this ever being used beyond current situation.

inventory.py
pushyamig: Done

inventory.py
pushyamig: This is the exception return from the API utils and it looks like it is a generic exception.

inventory.py
pushyamig: Done

inventory.py
pushyamig: Perhaps

inventory.py
pushyamig: Done

pushyamig: Done

pushyamig:
The reason why the recursion is there for paging through the data. I am holding the date found in a list if the code doesn't fetch the data in the first call and then only it goes into this recursion and during that time it it finds the value I use break to exit but due to recursion, it returns None. 

SO I felt I will hold the data in a list and will always return the first item(which will always the case). There may be a better way of doing this due to the time constraint I opted for this approach.

inventory.py
pushyamig: see my comment here https://github.com/tl-its-umich-edu/course-inventory/pull/9#issuecomment-600088985

inventory.py
pushyamig: See My comments here https://github.com/tl-its-umich-edu/course-inventory/pull/9#issuecomment-600088985

inventory.py
pushyamig: Done

inventory.py
```diff
@@ -105,47 +90,47 @@ def handle_request_if_failed(response):
...
         if event['event_type'] == 'published':
-            publish_date = event['created_at']
-            logger.info(f"Date for Workflow type :{df_row['course_workflow_state']} for course {course_id}")
+            publish_date.append(event['created_at'])
+            logger.info(f"Date for Workflow type {df_row['course_workflow_state']} for course {course_id}")
             break
-    if publish_date is None:
-        next_page_url = get_next_page_url(response)
+    if len(publish_date) == 0:
+        next_page_url = API_UTIL.get_next_page(response)
```
pushyamig: Changed the next page logic based on test case in the API utils https://github.com/tl-its-umich-edu/api-utils-python/blob/25aec8f0228cc426d463d5958428bbbc6ddd6d33/tests/test.py#L73.

I don't think the previous implementation of mine worked. But Now It will :)

inventory.py
```diff
@@ -172,10 +157,8 @@ def gather_course_info_for_account(account_id: int, term_id: int) -> Sequence[in
 
     course_df = pd.DataFrame(slim_course_dicts)
     course_df['course_warehouse_id'] = course_df['course_id'].map(lambda x: x + WAREHOUSE_INCREMENT)
-    course_date=None
-    course_df['published_date'] = course_df.apply(lambda x: get_course_publish_date(x, course_date), axis=1)
+    course_df['published_date'] = course_df.apply(lambda x: get_course_publish_date(x, []), axis=1)
```
pushyamig: The [] is support recursion

inventory.py
```diff
@@ -105,47 +90,47 @@ def handle_request_if_failed(response):
...
+            publish_date.append(event['created_at'])
+            logger.info(f"Date for Workflow type {df_row['course_workflow_state']} for course {course_id}")
             break
-    if publish_date is None:
-        next_page_url = get_next_page_url(response)
+    if len(publish_date) == 0:
+        next_page_url = API_UTIL.get_next_page(response)
         if next_page_url is not None:
-            get_course_publish_date(course_id, publish_date, next_page_url)
+            logger.info("Get date from next page")
```
pushyamig: I am adding a log statement here to see if more than 100-course log events for a course audit API. I doubt this is the case.

inventory.py
pushyamig: Also directly call the  API as like `canvasApiurl?page=34` etc is kind of discouraged by canvas and may not be supported in future, But it's not the problem to solve today

ssciolla: Can you point to where this is discoursaged? I did not know `api_utils_python` had paging. We can try to change my code to using that, especially if it will help reduce some duplication here.

inventory.py
ssciolla: What is a generic exception?

inventory.py
```diff
@@ -25,13 +25,17 @@
...
 
 UDW_CONN = psycopg2.connect(**ENV['UDW'])
 WAREHOUSE_INCREMENT = ENV['WAREHOUSE_INCREMENT']
 
 
 # Function(s)
 
+
+
+
```
ssciolla: Delete these?

inventory.py
pushyamig: @ssciolla I have sent you an email regarding that discussion. Based on that it only impacts enrollment API but Canvas [documentation](https://canvas.instructure.com/doc/api/file.pagination.html) suggests only to pull page links from link header

inventory.py
pushyamig: https://github.com/tl-its-umich-edu/api-utils-python/blob/master/umich_api/api_utils.py#L132

pushyamig:
Currently, I am trying to see if I can improve the performance of the process for pulling the published date. I did not see for getting a published date (with API response of 100 items ) I need to use the next page logic. I felt we might getting lot of course events in audit logs but it seems it is not the case So far(still testing). It is less likely that I am going with pagination/recussion logic
But I can say that on average it taking about 4-5 sec for each api call. I have used direct canvas API( instead of API directory for improving the performance ) it is also taking 4-5 sec.

I am currently testing if I just pull the published course list and try to run through fetching the published results instead of a bigger set with both un/published courses and later merge the data frame. I shouldn't see much difference but want to test it out. Here is the approch

```
course_df = pd.DataFrame(slim_course_dicts)
course_avail_df= course_df.loc[course_df.course_workflow_state =='available'].copy()
course_avail_df['published_date'] = course_avail_df['course_id'].map(lambda x: get_course_publish_date(x))
df = pd.merge(course_df, course_avail_df, on='course_id', how='left')
```
Next thing to try is Graphql

Issue #4, opened 2020-03-13
Change output to be a persistent database


ssciolla:
Right now the application just writes output to static CSV files. To simplify integration with the Tableau-side of things, it would make sense to use a persistent database. For simplicity, we should probably just completely drop the tables before rebuilding, but we should consider whether past states will need to be accessed.

PR #10, opened 2020-03-16
Create db module, change to inserting data into MySQL database (#4)

ssciolla:
This PR introduces the local `db` module for creating and managing the MySQL database schema and then changes the output strategy to insert data in an arbitrary MySQL database. Other changes were made to generally re-factor, limit the enrollment record results, and make the application more configurable. A task list with more detail will be provided below. This PR aims to resolve issue #4.

### Task list

- [x] Create `db` module with `DBCreator` class and `tables.py` file containing the MySQL create statements.
- [x] Create root level `create_db.py` script that uses the `db` module to create the database schema.
- [x] Break up UDW data pulling function into separate functions for user and enrollment data pulling.
- [x] Re-name columns to eliminate redundancy in database schema.
- [x] Add code that checks for nonexistent user ids in enrollment records, and then only keeps those with valid IDs.
- [x] Add `mysqlclient` to `requirements.txt` and add block to `Dockerfile` installing `default-libmysqlclient-dev` and other packages (based this on other applications' `Dockerfile`s.
- [x] Make `ACCOUNT_ID` and `INVENTORY_DB` (MySQL database) configurable through `env.json`, and tweak a couple other environment variable names.
- [x] Add `CREATE_CSVS` environment variable setting that makes CSV generation optional.

### DIFF
--- a/db/create_db.py
+++ b/db/create_db.py
@@ -1,0 +1,85 @@
+# standard libraries
+import logging, json, os
+from datetime import datetime
+from typing import Dict, Sequence
+
+# third-party libraries
+import pandas as pd
+from sqlalchemy import create_engine
+
+# local libraries
+from db.tables import tables
+
+
+# Initializing settings and global variables
+
+logger = logging.getLogger(__name__)
+
+try:
+    with open(os.path.join('config', 'env.json')) as env_file:
+        ENV = json.loads(env_file.read())
+except FileNotFoundError:
+    logger.error('Configuration file could not be found; please add env.json to the config directory.')
+
+DB_PARAMS = ENV['INVENTORY_MYSQL']
+
+conn_str = (
+    'mysql' +
+    f"://{DB_PARAMS['user']}" +
+    f":{DB_PARAMS['password']}" +
+    f"@{DB_PARAMS['host']}" +
+    f":{DB_PARAMS['port']}" +
+    f"/{DB_PARAMS['dbname']}?charset=utf8"
+)
+
+logger.info(conn_str)
+
+MYSQL_ENGINE = create_engine(conn_str)
+
+
+# Class(es)
+
+class DBCreator:
+
+    def __init__(self, db_name: str, table_dicts: Sequence[Dict[str, str]]):
+        self.conn = None
+        self.db_name = db_name
+        self.tables = table_dicts
+
+    def set_up(self):
+        logger.debug('set_up')
+        self.conn = MYSQL_ENGINE.connect()
+
+    def tear_down(self):
+        logger.debug('tear_down')
+        self.conn.close()
+
+    def drop_tables(self) -> None:
+        logger.debug('drop_tables')
+        self.conn.execute('SET FOREIGN_KEY_CHECKS=0;')
+        table_names = [table['name'] for table in tables]
+        drop_statement = f'DROP TABLE IF EXISTS {", ".join(table_names)};'
+        self.conn.execute(drop_statement)
+        self.conn.execute('SET FOREIGN_KEY_CHECKS=1;')
+
+    def create_tables(self) -> None:
+        logger.debug('create_tables')
+        for table_dict in self.tables:
+            table_name = table_dict["name"]
+            logger.debug(f'Table Name: {table_name}')
+            self.conn.execute(table_dict['statement'])
+            logger.info(f'Created table {table_name} in {self.db_name}')
+
+    def set_up_database(self) -> None:
+        self.set_up()
+        self.drop_tables()
+        self.create_tables()
+        self.tear_down()
+
+
+# Main Program
+
+if __name__ == '__main__':
+    logging.basicConfig(level=ENV.get('LOG_LEVEL', 'DEBUG'))
+    db_creator_obj = DBCreator('course_inventory', tables)
+    db_creator_obj.set_up_database()
--- a/db/create_db.py
+++ b/db/create_db.py
@@ -21,7 +21,7 @@ try:
 except FileNotFoundError:
     logger.error('Configuration file could not be found; please add env.json to the config directory.')
 
-DB_PARAMS = ENV['INVENTORY_MYSQL']
+DB_PARAMS = ENV['INVENTORY_DB']
 
 conn_str = (
     'mysql' +
--- a/db/create_db.py
+++ b/db/create_db.py
@@ -59,10 +59,20 @@ class DBCreator:
         self.conn.execute(drop_statement)
         self.conn.execute('SET FOREIGN_KEY_CHECKS=1;')
 
+    def drop_records(self) -> None:
+        logger.debug('drop_records')
+        self.conn.execute('SET FOREIGN_KEY_CHECKS=0;')
+        for table_dict in self.tables:
+            table_name = table_dict['name']
+            logger.debug(f'Table Name: {table_name}')
+            self.conn.execute(f'DELETE FROM {table_name};')
+            logger.info(f'Dropped records in {table_name} in {self.db_name}')
+        self.conn.execute('SET FOREIGN_KEY_CHECKS=1;')
+
     def create_tables(self) -> None:
         logger.debug('create_tables')
         for table_dict in self.tables:
-            table_name = table_dict["name"]
+            table_name = table_dict['name']
             logger.debug(f'Table Name: {table_name}')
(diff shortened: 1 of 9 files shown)


inventory.py
```diff
@@ -93,57 +99,105 @@ def gather_course_info_for_account(account_id: int, term_id: int) -> Sequence[in
...
+    user_ids = user_df['warehouse_id'].drop_duplicates().to_list()
+
+    def check_if_valid_user_id(id: int) -> bool:
+        if id in user_ids:
+            return True
+        else:
+            return False
+
+    enrollment_df['valid_id'] = enrollment_df['user_id'].map(check_if_valid_user_id)
+    enrollment_df = enrollment_df[(enrollment_df['valid_id'])].drop(columns=['valid_id'])
```
ssciolla: This operation is necessary, but it is time intensive (a little less than three minutes, I think). I looked for alternatives like including an `INNER JOIN` in the UDW enrollment query, but I wasn't able to get anything to work. Any ideas?

inventory.py
```diff
@@ -93,57 +99,109 @@ def gather_course_info_for_account(account_id: int, term_id: int) -> Sequence[in
...
+        return True
+    else:
+        return False
+
+
+def run_course_inventory() -> None:
+    start = datetime.now()
+
+    # Gather course data
+    course_df = gather_course_info_for_account(1, ENV['TERM_ID'])
```
zqian: account id value could be set in the env.json file. It may have different value than the root account.

inventory.py
```diff
@@ -93,57 +99,109 @@ def gather_course_info_for_account(account_id: int, term_id: int) -> Sequence[in
...
+        # Generate CSV Output
+        course_df.to_csv(os.path.join('data', 'course.csv'), index=False)
+        logger.info('Course data was written to data/course.csv')
+        user_df.to_csv(os.path.join('data', 'user_df.csv'), index=False)
+        logger.info('User data was written to data/user.csv')
+        enrollment_df.to_csv(os.path.join('data', 'enrollment.csv'), index=False)
+        logger.info('Enrollment data was written to data/enrollment.csv')
+
+    # Reset database
+    db_creator_obj = DBCreator('course_inventory', tables)
```
zqian: so the database will be dropped and recreated every time the script runs?

inventory.py
```diff
@@ -93,57 +99,109 @@ def gather_course_info_for_account(account_id: int, term_id: int) -> Sequence[in
...
+def pull_enrollment_data_from_udw(course_ids) -> pd.DataFrame:
+    courses_string = ','.join([str(course_id) for course_id in course_ids])
     enrollment_query = f'''
-        SELECT e.id AS enrollment_id,
-               e.canvas_id AS enrollment_canvas_id,
-               e.user_id AS enrollment_user_id,
-               e.course_section_id AS enrollment_course_section_id,
-               e.course_id AS enrollment_course_id,
-               e.workflow_state AS enrollment_workflow_state,
+        SELECT e.id AS warehouse_id,
```
zqian: we may only want to select student enrollment with e.workflow_state='active'

inventory.py
```diff
@@ -93,57 +99,109 @@ def gather_course_info_for_account(account_id: int, term_id: int) -> Sequence[in
...
+        logger.info('Enrollment data was written to data/enrollment.csv')
+
+    # Reset database
+    db_creator_obj = DBCreator('course_inventory', tables)
+    db_creator_obj.set_up_database()
+
+    # Insert gathered data
+    logger.info('Inserting course data')
+    course_df.to_sql('course', MYSQL_ENGINE, if_exists='append', index=False)
+    logger.info('Inserting user data')
```
zqian: Can we put the dataframe size information into the log message? Same for the csv output logs. Thanks!

inventory.py
ssciolla: I'll do this!

db/create_db.py
```diff
@@ -0,0 +1,85 @@
...
+    def tear_down(self):
+        logger.debug('tear_down')
+        self.conn.close()
+
+    def drop_tables(self) -> None:
+        logger.debug('drop_tables')
+        self.conn.execute('SET FOREIGN_KEY_CHECKS=0;')
+        table_names = [table['name'] for table in tables]
+        drop_statement = f'DROP TABLE IF EXISTS {", ".join(table_names)};'
+        self.conn.execute(drop_statement)
```
ssciolla: Try just dropping the records, not the tables

inventory.py
ssciolla: Let's check with @jennlove-um about this.

jennlove-um: Selecting onlyl "active" students should be okay.  The only other state that I was considering was "invited," but until students accept the invitation, they aren't really enrolled.  So let's go with workflow_state='active' as suggested.

ssciolla: I'm seeing these values in the enrollment table for `workflow_state`.
```| active           |
| deleted          |
| completed        |
| invited          |
| inactive         |
| rejected         |
| creation_pending```

ssciolla: Which of these should we keep?

jennlove-um: We should keep 'active' because for this report we only care about students who are actively enrolled in the course.  None of those other statuses should matter for this.

inventory.py
jonespm: Yeah, I suggested to look into a migration package like [Yoyo](https://pypi.org/project/yoyo-migrations/) or [Alembic](https://alembic.sqlalchemy.org/en/latest/) but maybe that was going to take too much time to implement. Seems useful for the future if data needs to be retained.

ssciolla: Yeah, I wasn't trying to ignore your suggestion, just trying to get this finished. We can explore this in the future.

I don't think data needs to be retained between runs, but I could be convinced otherwise.

inventory.py
ssciolla: Okay, I will try this out.

inventory.py
jonespm: Yeah I know how long it takes to do some of this stuff and wasn't sure the entire usage of this project. I haven't used any python migration other than the Django ORM and Flake-Migrate (that works like Alembic) but would be interested in learning more. I know the lack of migrations or poorly managed migrations can often cause headaches if the project runs for awhile.

inventory.py
ssciolla: This should be resolved now.

inventory.py
ssciolla: This should be taken care of by [this commit](https://github.com/tl-its-umich-edu/course-inventory/pull/10/commits/0332accecb4912348fbc427fef75f2cfbe153c19).

db/create_db.py
ssciolla: Done, as of [this commit](https://github.com/tl-its-umich-edu/course-inventory/pull/10/commits/4b31ac76a1db99b89e16ddafbc0069e2267c288c).

inventory.py
ssciolla: I am now filtering on the 'active' `workflow_state`, as of [this commit](https://github.com/tl-its-umich-edu/course-inventory/pull/10/commits/1f312a45da35623f71e54a63c0566b8a34447717).

inventory.py
ssciolla: I had hoped that this would become unnecessary once we began filtering on the enrollment `workflow_state` being 'active', but I'm still getting the error, with small numbers of user IDs not being present in the `user` table.

inventory.py
ssciolla: See [this commit](https://github.com/tl-its-umich-edu/course-inventory/pull/10/commits/9ab08d5cf186428d81a678fa441cd13ae57146f4).

ssciolla:
FYI, @pushyamig pointed out some errors with the previous `create_db.py` file, so I investigated and ended up refactoring a bit. The `db` module now only contains a `db_creator` file with the `DBCreator` class and the `tables.py` with the `CREATE TABLE` statements. That module is now used by a root-level `create_db.py`.

zqian:
I have problem using docker version to connect with local MySQL instance: "ERROR 2002 (HY000): Can't connect to local MySQL server through socket '/var/run/mysqld/mysqld.sock'"

Running the virtual environment version has no problem, except when the database is empty with no tables, the drop_tables call generated error. I will test it again today.

Issue #11, opened 2020-03-17
Use get_next_page method from api_utils_python for paging


ssciolla:
In discussing PR #9 with @pushyamig, I learned that the `ApiUtil` class in `api_utils_python` also has a `get_next_page` method that helps with paging. @pushyamig also pointed out that the way I originally implemented paging (just using `page=x` in the `payload`) is not recommended. To follow best practices and perhaps make the functions I wrote more easy to integrate with PR #9, I'll try to make this switch.

PR #12, opened 2020-03-17
Use ApiUtil.get_next_page for paging (#11)

ssciolla:
This PR makes changes to the Canvas API course data fetching so that it employs the recommended strategy for pagination, utilizing the link headers. The code now uses the `get_next_page` method on the `ApiUtil` class to get the proper parameters for the next request. To enable this change, I modified `make_request_using_api_utils` so that it returns a `requests.Response` object rather than the parsed data. This PR aims to resolve issue #11.

### DIFF
--- a/inventory.py
+++ b/inventory.py
@@ -53,7 +53,7 @@ def make_request_using_api_utils(url: str, params: Dict[str, Union[str, int]] = 
                 logger.warning('JSONDecodeError encountered')
                 logger.info('Beginning next attempt')
 
-    logger.error('The maximum number of reqeust attempts was reached')
+    logger.error('The maximum number of request attempts was reached')
     return response
 
 
--- a/inventory.py
+++ b/inventory.py
@@ -6,6 +6,7 @@ from typing import Dict, Sequence, Union
 # third-party libraries
 import pandas as pd
 import psycopg2
+from requests import Response
 from umich_api.api_utils import ApiUtil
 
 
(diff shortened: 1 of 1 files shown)


inventory.py
```diff
@@ -71,23 +72,30 @@ def slim_down_course_data(course_data: Sequence[Dict]) -> Sequence[Dict]:
...
-        logger.info(f"Course Page Number: {params['page']}")
-        all_course_data = make_request_using_api_utils(f'{API_SCOPE_PREFIX}/{url_ending}', params)
-        if len(all_course_data) > 0:
+        next_params = API_UTIL.get_next_page(response)
+        if next_params:
+            page_num += 1
+            logger.info(f'Course Page Number: {page_num}')
+            response = make_request_using_api_utils(url_ending_with_scope, next_params)
+            all_course_data = json.loads(response.text)
             slim_course_dicts += slim_down_course_data(all_course_data)
```
zqian: The coding within line 85-88 and line 95-98 are very similar. Could them be refactored into a single function?

ssciolla: I could, yes. However, it seems it wouldn't be terribly useful outside of this function, since it's using this `slim_down_course_data` function.

Part of the reason for these changes though was to generalize the existing request function so that it only returned a response (thus making it necessary to do the `json.loads(response.text)` operation). Hopefully this means we can remove some of the duplicated functionality @pushyamig had created for her PR.

zqian: sounds good to me! I will close this request then.

Issue #13, opened 2020-03-18
Use yoyo-migrations to version control, manage the DB


ssciolla:
Though it doesn't make sense to use Django, since Tableau is the FE, it would still be nice to have some database migration functionality. Seems like [`yoyo-migrations`](https://pypi.org/project/yoyo-migrations/#description) might be a simple way to do this.

PR #16, opened 2020-03-19
Allow for ENV_PATH variable for OpenShift integration, add documentation (#1, #3)

ssciolla:
This PR tweaks how the `env.json` configuration file is loaded, moving the file's expected location to a `secrets` sub-directory inside the `config` directory and allowing for an `ENV_PATH` environment variable to be set in OpenShift to set the path (allowing for different file names, primarily). It also adds an application overview, configuration instructions, and OpenShift deployment details to `README.md`, as well as makes general formatting changes to the file. This PR aims to aims to resolve issue #1 and finish resolving issue #3.

PR #17, opened 2020-03-20
Manage database with yoyo-migrations (#13)

ssciolla:
This PR modifies the `db` module  -- and uses of it in `create_db.py` and `inventory.py` -- to integrate the [`yoyo-migrations`](https://ollycope.com/software/yoyo/latest/) library to provide to version control and make incremental changes to the database schema. A task list will be provided below. This PR aims to resolve issue #13.

- [x] Add `yoyo-migrations` to `requirements.txt`
- [x] Remove `tables.py` and create replacement `migrations` directory with `0001.init_inventory_db.py`.
- [x] Refactor `db_creator.py`.
   - [x] Simplify constructor parameters to accept all database parameters
   - [x] Create `conn_str` and `engine` inside `__init__` method.
   - [x] Replace `create_tables` with `migrate` method.
   - [x] Remove `drop_tables` method.
- [x] Update references to `DBCreator` and `db` module in `create_db.py` and `inventory.py`.


pushyamig:
How does this work? Do I need to run any migration command to migrate?

ssciolla:
> How does this work? Do I need to run any migration command to migrate?

The `migrate` method in `DBCreator` should handle running the migrations. If you're using `virtualenv`, I'd reset your database first using `create_db.py`, but going forward, there is code in `inventory.py` that will drop records and apply any outstanding migrations before trying to insert.

There are also command line operations for doing migrations; see [the documentation](https://ollycope.com/software/yoyo/latest/).

db/yoyo.ini
```diff
@@ -0,0 +1,9 @@
+# List of migration source directories. "%(here)s" is expanded to the
```
jonespm: It doesn't look like this file is used when you're using it in python mode. This is just here so we can create a new migration or run "yoyo" on the CLI? I'm not sure. I'm not sure if that's something we'd ever do with this project though.

pushyamig:
Ran the app and this is creating the table and adding data to it. Read me should be updated with couple of things
1. The MySQL DB set up, creating the DATABASE 'course_inventory'  is a pre-req (That's what I did), user and privileges  etc
2. When there is a model change required README show to tell how to generate the migration file for record-keeping.

db/yoyo.ini
ssciolla: Yeah, looks like you are correct. Deleting it doesn't change anything. I thought at first the `verbosity` setting was doing something, but I think with the Python mode, we just use `LOG_LEVEL`. I'll remove this for now, I think we can just generate new migration files manually and test.

jonespm:
@pushyamig 
1. @ssciolla  has a docker-compose coming up that will create the database if you use the MySQL in the VM. 

2. I do think we should mention about migration files, but I think it's similar to Django where you just add a new sql or python file to the migrations directory with a __depends__ in them. The difference is that it's SQL rather than Django style code.  Maybe just link to https://ollycope.com/software/yoyo/latest/#migration-files

ssciolla:
@pushyamig and @jonespm , I will merge this add some details about MySQL and migrations as part of PR #16.

PR #18, opened 2020-03-20
I8 get course published date

### LINKED ISSUE #8: getting the published date for a course

pushyamig:
Fixes #8 

1. I have shared a log file with you all when ran for the full process for getting course, user, enrollment data with `15 worker threads` and `page_page=50` results

I am using `[requests-futures](https://github.com/ross/requests-futures)` library for shorting the processing time for pull published date info from course_audit logs.  
By default, Library uses a ThreadPoolExecutor is created with 8 workers time took was ~1hr. I tried to increase worker threads to 15 took about ~40 minutes and I did not see any 403 (Forbidden for rate-limiting exceeding) error.  I haven't tested what happens if I used >15 workers. 

A little bit of background:
I was using API directory for pulling published date info. With API directory I started with sequential approach(one call after the other) it took about 7hrs time, then I opted for using Python thread and Queues approach this was not effective as well since with large process something the way the Operating system handles the process will finish but never exits to create a CSV file. That’s the reason why I opted for the different library called `requests-futures`. I need to use Canvas ADMIN token and using direct canvas AP with this approach.


config/env_blank.json
```diff
@@ -2,6 +2,8 @@
   "LOG_LEVEL": "DEBUG",
   "CANVAS_ACCOUNT_ID": 1,
   "CANVAS_TERM_ID": 164,
+  "CANVAS_TOKEN":"",
+  "CANVAS_URL": "",
```
pushyamig: Since I am using the different library async calling and it is handling all the Http operation so I  need to use direct canvas API;s

db/migrations/0002.add_col_published_date.py
```diff
@@ -0,0 +1,8 @@
+#
+# file: migrations/0002.add_col_published_date.py
+#
+from yoyo import step
+
+__depends__ = {'0001.init_inventory_db'}
+
+step("ALTER TABLE course ADD column published_date VARCHAR(25) AFTER created_at")
```
pushyamig: Thinks were simple with migration, I am Q though why is this VARCHAR instead of datetime, I just followed what was for `created_Date`.

inventory.py
```diff
@@ -34,12 +37,17 @@
...
 MAX_REQ_ATTEMPTS = ENV['MAX_REQ_ATTEMPTS']
+CANVAS_TOKEN = ENV['CANVAS_TOKEN']
+CANVAS_URL = ENV['CANVAS_URL']
 
 UDW_CONN = psycopg2.connect(**ENV['UDW'])
 WAREHOUSE_INCREMENT = ENV['WAREHOUSE_INCREMENT']
 
 CREATE_CSVS = ENV.get('CREATE_CSVS', False)
 INVENTORY_DB = ENV['INVENTORY_DB']
+published_course_date = {}
```
pushyamig: This is holding all the published date info from async call.

inventory.py
```diff
@@ -34,12 +37,17 @@
...
+published_course_next_page_list = []
```
pushyamig: This holds the pagination URL to query next for courses that don't get published_date from the first 100 results. So far only 2 courses from 5k published courses list I had to loop through

pushyamig: Since library controls when to fetch the next results I need to hold it somewhere outside of the function call routine

db/migrations/0002.add_col_published_date.py
ssciolla: Yeah, it should probably be `TIMESTAMP` or one of the other time-related data types. I encountered problems when trying to insert the data into the database, so I just forged ahead with VARCHAR for the moment. If you can use something other than VARCHAR, please do. I can work on getting the other one changed back. Thanks for reminding me.

inventory.py
```diff
@@ -227,8 +328,9 @@ def run_course_inventory() -> None:
     enrollment_df.to_sql('enrollment', db_creator_obj.engine, if_exists='append', index=False)
     logger.info(f'Inserted data into enrollment table in {db_creator_obj.db_name}')
 
-    delta = datetime.now() - start
-    logger.info(f'Duration of run: {delta.total_seconds()}')
+    delta = time.time() - start
+    str_time = time.strftime("%H:%M:%S", time.gmtime(delta))
+    logger.info(f'Duration of run: {str_time}')
```
pushyamig: This is nicely printing the time taken for all the call like this `INFO:__main__:Duration of run: 01:06:46`

config/env_blank.json
ssciolla: We should probably talk about whether we can use a mixed approach. I've heard from T&L leadership that personal access tokens are only okay in the short-term.

inventory.py
ssciolla: Is there a reason we can't use `datetime` for this?

pushyamig:
This PR is working but I opened as draft PR I just want to check a few things before I open it up. I just want all have a peek of what I have been doing and going back and forth in slack with my questions.

config/env_blank.json
pushyamig: We need to have some sort of admin token created for this. This PR using a different library as http client is a quick whip to reduces the timing on the process. I don't know if we need to think a long time solution

db/migrations/0002.add_col_published_date.py
pushyamig: It's all boils down to how tableau users wants the data to be like. it will make sense both date to be same data type otherwise on tableau end they might need to do some tableau calculation or tweaks to adjust to same data type.

inventory.py
pushyamig: I wanted in a format of H:M:S so that based on the time take it to adjust itself to seconds, minutes etc. I did not see that with existing approach was doing it so I changed it. I don't have a specific reason for why I chose this

config/env_blank.json
ssciolla: I'm just communicating what was told to me. I agree that in the short term the personal access token is an easy solution (it is also for GraphQL). I will say, in terms of what I'm seeing and hearing, this is likely going to be a long-term project.

db/migrations/0002.add_col_published_date.py
ssciolla: Agreed, it should be consistent. You can keep it the same data type for now, and I can work on improving both later.

inventory.py
ssciolla: Not sure the best way to do this. It looks like you're passing seconds into `gmtime`, which is really calculating the time since the epoch began, and then you're just looking at the hours, minutes, and seconds. That's probably fine? I'm not married to my approach either; the `timedelta` objects seemed appropriate here, but you'd have to do a little math to get the hours and minutes. If you decided to stick to your approach, please remove the `datetime` import.

config/env_blank.json
```diff
@@ -10,6 +12,7 @@
   "MAX_REQ_ATTEMPTS": 3,
   "WAREHOUSE_INCREMENT": 0,
   "CREATE_CSVS": false,
+  "PUBLISHED_DATE_WORKERS": 8,
```
pushyamig: I increased up to 15 workers time take to run got decreased without errors(403). I might need to test what is the breaking point to start seeing 403's

db/migrations/0002.add_col_published_date.py
```diff
@@ -0,0 +1,8 @@
+#
+# file: migrations/0002.add_col_published_date.py
+#
+from yoyo import step
+
+__depends__ = {'0001.init_inventory_db'}
+
+step("ALTER TABLE course ADD column published_at DATETIME NULL AFTER created_at")
```
pushyamig: The `published_at` is datatime and underneath that, it stores the UTC date. I felt data time serves our purpose

db/migrations/0003.created_date_datatype_change.py
```diff
@@ -0,0 +1,8 @@
+#
+# file: migrations/0003.created_date_datatype_change.py
+#
+from yoyo import step
+
+__depends__ = {'0002.add_col_published_date'}
+
+step("ALTER TABLE course MODIFY column created_at DATETIME NULL")
```
pushyamig: Changes the `created_at` data type to be consistence with `published_at` date

inventory.py
```diff
@@ -25,7 +28,10 @@
 except FileNotFoundError:
     logger.error('Configuration file could not be found; please add env.json to the config directory.')
 
-logging.basicConfig(level=ENV.get('LOG_LEVEL', 'DEBUG'))
+logging.basicConfig(level=ENV.get('LOG_LEVEL', 'DEBUG'),
+                    format='%(asctime)s - %(name)s - %(levelname)s - %(message)s',
+                    handlers=[RotatingFileHandler('debug.log', mode='w', maxBytes=10485760,backupCount=5),
+                              logging.StreamHandler()])
```
pushyamig: The reason I was using a `FileHandler` is I want to store logs for reviewing since the calls are async (the library is handling it ) I need to go back and read the logs to make sense. 
This shouldn't affect Openshift logging and shouldn't throw any error when Openshift disallow for creating a new file with `debug.log`. We can watch out for it this

inventory.py
```diff
@@ -167,14 +180,115 @@ def check_if_valid_user_id(id: int, user_ids: Sequence[int]) -> bool:
...
+            responses = []
+            for next_page_link in next_page_links:
+                response = session.get(next_page_link, headers=headers)
+                published_course_next_page_list.remove(next_page_link)
+                responses.append(response)
+        else:
+            logger.info("Initial Round of Fetching course published date")
+            responses = [
+                session.get(f'{CANVAS_URL}/api/v1/audit/course/courses/{course_id}?per_page=100', headers=headers)
+                for course_id in course_ids]
```
pushyamig: The library fetches all the responses based on the worker and processes it as a response received. If a response of the first 100 results doesn't have data then we create a pagination course list and once this process completes it moves on to fetching data from the pagination list as inline 256-259. The debug.log that i shared with you should reflect that.

I intentionally limited the page size to 50 to see how the paging logic goes and I see data being correctly pulled

inventory.py
```diff
@@ -167,14 +180,115 @@ def check_if_valid_user_id(id: int, user_ids: Sequence[int]) -> bool:
...
     course_df = gather_course_info_for_account(ACCOUNT_ID, TERM_ID)
+    course_available_df= course_df.loc[course_df.workflow_state == 'available'].copy()
+    course_available_ids = course_available_df['canvas_id'].to_list()
+    get_published_course_date(course_available_ids)
+    df1 = pd.DataFrame(published_course_date.items(), columns=['canvas_id','published_at'])
+    course_complete_df = pd.merge(course_df, df1, on='canvas_id', how='left')
+    course_complete_df['created_at'] = pd.to_datetime(course_complete_df['created_at'], format="%Y-%m-%dT%H:%M:%SZ",
+                                                      errors='coerce')
+    course_complete_df['published_at'] = pd.to_datetime(course_complete_df['published_at'], format="%Y-%m-%dT%H:%M:%SZ",
+                                                        errors='coerce')
```
pushyamig: The code in Line 285-287 was done so that inserts to DB with null is only allowed when `publised_at` and `created_at` columns are DateTime instead of null. `errors='coerce'` it is enforcing `Nat` when None in those columns

inventory.py
pushyamig: Done

db/migrations/0003.created_date_datatype_change.py
ssciolla: Yeah, this change makes sense to me. I was also looking at `DATETIME`. I'm OK including this here, though I thought we were talking about doing this as a separate issue.

inventory.py
ssciolla: Does this need to be part of the application? Could you just use it for testing purposes?

inventory.py
```diff
@@ -34,16 +40,23 @@
 SUBSCRIPTION_NAME = ENV['API_SUBSCRIPTION_NAME']
 API_SCOPE_PREFIX = ENV['API_SCOPE_PREFIX']
 MAX_REQ_ATTEMPTS = ENV['MAX_REQ_ATTEMPTS']
+CANVAS_TOKEN = ENV['CANVAS_TOKEN']
+CANVAS_URL = ENV['CANVAS_URL']
+PUBLISHED_DATE_WORKERS = ENV.get('PUBLISHED_DATE_WORKERS',8)
```
ssciolla: Could this be generalized in case we end up using asynchronous calls for something else? Maybe `ASYNC_WORKERS`?

inventory.py
```diff
@@ -167,14 +180,115 @@ def check_if_valid_user_id(id: int, user_ids: Sequence[int]) -> bool:
...
 def run_course_inventory() -> None:
-    start = datetime.now()
+    start = time.time()
 
     # Gather course data
     course_df = gather_course_info_for_account(ACCOUNT_ID, TERM_ID)
+    course_available_df= course_df.loc[course_df.workflow_state == 'available'].copy()
+    course_available_ids = course_available_df['canvas_id'].to_list()
+    get_published_course_date(course_available_ids)
+    df1 = pd.DataFrame(published_course_date.items(), columns=['canvas_id','published_at'])
```
ssciolla: Please use a better variable name than `df1`.

inventory.py
```diff
@@ -167,14 +180,115 @@ def check_if_valid_user_id(id: int, user_ids: Sequence[int]) -> bool:
...
+    course_complete_df = pd.merge(course_df, df1, on='canvas_id', how='left')
+    course_complete_df['created_at'] = pd.to_datetime(course_complete_df['created_at'], format="%Y-%m-%dT%H:%M:%SZ",
+                                                      errors='coerce')
+    course_complete_df['published_at'] = pd.to_datetime(course_complete_df['published_at'], format="%Y-%m-%dT%H:%M:%SZ",
+                                                        errors='coerce')
 
     # Gather enrollment data
-    udw_course_ids = course_df['warehouse_id'].to_list()
+    udw_course_ids = course_complete_df['warehouse_id'].to_list()
```
ssciolla: Could we re-assign `course_complete_df` to `course_df` somewhere above so you we don't need to change this and the other modifications below?

config/env_blank.json
```diff
@@ -2,6 +2,8 @@
   "LOG_LEVEL": "DEBUG",
   "CANVAS_ACCOUNT_ID": 1,
   "CANVAS_TERM_ID": 164,
+  "CANVAS_TOKEN":"",
```
ssciolla:
```suggestion
  "CANVAS_TOKEN": "",
```

inventory.py
ssciolla:
```suggestion
PUBLISHED_DATE_WORKERS = ENV.get('PUBLISHED_DATE_WORKERS', 8)
```

inventory.py
```diff
@@ -34,16 +40,23 @@
...
 CREATE_CSVS = ENV.get('CREATE_CSVS', False)
 INVENTORY_DB = ENV['INVENTORY_DB']
+published_course_date = {}
+published_course_next_page_list = []
+
 
 
 # Function(s)
 
+
```
ssciolla: I don't think these newlines are needed.

inventory.py
```diff
@@ -167,14 +180,115 @@ def check_if_valid_user_id(id: int, user_ids: Sequence[int]) -> bool:
...
+    if not results:
+        logging.debug('The api call do not have Link headers')
+        return None
+
+    for page in results:
+        if 'next' in page:
+            url_ = results['next']['url']
+            published_course_next_page_list.append(url_)
+            logger.info(f"Pagination size {len(published_course_next_page_list)} for published_at date")
+            break
```
ssciolla: Can you just check for "next" in `result.keys()` rather than looping here?

inventory.py
ssciolla: This is cool! I would like to see the `to_datetime` calls spaced out like below, but not required.

```
course_complete_df['published_at'] = pd.to_datetime(
   course_complete_df['published_at'], 
   format="%Y-%m-%dT%H:%M:%SZ",
   errors='coerce'
)
```

inventory.py
```diff
@@ -167,14 +180,115 @@ def check_if_valid_user_id(id: int, user_ids: Sequence[int]) -> bool:
...
+        get_published_course_date(course_ids, published_course_next_page_list)
+
+
 def run_course_inventory() -> None:
-    start = datetime.now()
+    start = time.time()
 
     # Gather course data
     course_df = gather_course_info_for_account(ACCOUNT_ID, TERM_ID)
+    course_available_df= course_df.loc[course_df.workflow_state == 'available'].copy()
```
ssciolla:
```suggestion
    course_available_df = course_df.loc[course_df.workflow_state == 'available'].copy()
```

inventory.py
ssciolla: As we talked about over BlueJeans, I would rather see these defined within the function scope of some wrapper function around the recursive calls, or you could create a class and have these be instance variables and your functions can be methods. Using global variables and functions that don't return anything (but are really modifying these globals) seems confusing (at least to me) and probably hard to maintain. My two cents.

inventory.py
ssciolla: Or maybe `NUM_ASYNC_WORKERS` would be more clear?

config/env_blank.json
zqian: Let's use the token for now. I hope we will see the /audit/account API release available soon. We can switch back to the API directory approach by then.

config/env_blank.json
zqian: This is OK for now. We do not need spent more time to fine-tune the parameters further.

inventory.py
zqian: sounds like this won't work in OpenShift environment. If it serves purpose only for local testing, let's remove this from the pull request.

inventory.py
pushyamig: @ssciolla I want to do it as separate PR since this approach is working and when this gets merged so that Tableau can pull directly from DB instead of csv's as it already started doing it. I might play around it a bit with refactoring. I don't know if I can completely avoid both the globals.

inventory.py
pushyamig: Removed the handler

inventory.py
pushyamig: Done

inventory.py
pushyamig: Done

inventory.py
pushyamig: Done

inventory.py
pushyamig: Done, I was hesitant to do this since you are merging  2 df and assigning to one of the df. But pandas did not complain. So I change it now

config/env_blank.json
pushyamig: Done

inventory.py
pushyamig: DOne

inventory.py
pushyamig: Done

inventory.py
pushyamig: Done

inventory.py
ssciolla: Yeah, it's a little strange, but I think this works.

inventory.py
```diff
@@ -167,11 +175,112 @@ def check_if_valid_user_id(id: int, user_ids: Sequence[int]) -> bool:
...
+            logger.info("Going through Next page URL set")
+            responses = []
+            for next_page_link in next_page_links:
+                response = session.get(next_page_link, headers=headers)
+                published_course_next_page_list.remove(next_page_link)
+                responses.append(response)
+        else:
+            logger.info("Initial Round of Fetching course published date")
+            responses = [
+                session.get(f'{CANVAS_URL}/api/v1/audit/course/courses/{course_id}?per_page=100', headers=headers)
```
ssciolla: Does `session.get` accept query parameters? I'm wondering why we're hardcoding the `per_page` parameter here.

inventory.py
```diff
@@ -1,14 +1,17 @@
...
 from json.decoder import JSONDecodeError
 from typing import Dict, Sequence, Union
+import time
 
 # third-party libraries
 import pandas as pd
 import psycopg2
 from requests import Response
 from umich_api.api_utils import ApiUtil
+from published_date.published_date import FetchPublishedDate
```
ssciolla: Can you put this in under local libraries with `db_creator`?

published_date/published_date.py
```diff
@@ -0,0 +1,109 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class FetchPublishedDate:
+
+    def __init__(self, canvas_url, canvas_token, workers, canvas_ids):
+        self.canvas_url = canvas_url,
+        self.canvas_token = canvas_token,
+        self.canvas_ids = canvas_ids
```
ssciolla: May want to call these `course_ids` or `canvas_course_ids` for clarity.

published_date/published_date.py
```diff
@@ -0,0 +1,109 @@
...
+        self.workers = workers
```
ssciolla: `num_workers`? `workers` makes me thing they are objects of some kind.

inventory.py
pushyamig: I don't think that makes sense. Since this is simply pulling data and no DB transactions

published_date/published_date.py
pushyamig: I just synced with naming conventions used in the `slim_down_course_data`

inventory.py
ssciolla: It's a local library though, not a third-party one?

published_date/published_date.py
ssciolla: I'm just concerned that this class makes no reference to a course entity in `__init__.py` or its name. Not a huge deal in the short term.

published_date/published_date.py
```diff
@@ -0,0 +1,109 @@
...
+                logger.info("Going through Next page URL set")
+                responses = []
+                for next_page_link in next_page_links:
+                    response = session.get(next_page_link, headers=headers)
+                    self.published_course_next_page_list.remove(next_page_link)
+                    responses.append(response)
+            else:
+                logger.info("Initial Round of Fetching course published date")
+                responses = [session.get(f'{self.canvas_url[0]}/api/v1/audit/course/courses/{course_id}?per_page=100', headers=headers)
+                for course_id in course_ids]
```
ssciolla: I'd probably refactor this for readability, probably making the url a variable on the previous line and then using it on this one.

published_date/published_date.py
```diff
@@ -0,0 +1,109 @@
...
+
+        for event in events:
+            if event['event_type'] == 'published':
+                course_id = event['links']['course']
+                published_date_found = True
+                self.published_course_date.update({course_id: event['created_at']})
+                logger.info(f"Published Date {event['created_at']} for course {course_id}")
+                break
+        if not published_date_found:
+            self.get_next_page_url(response)
```
ssciolla: So if there were multiple `published` events (i.e. an instructor published and un-published the course multiple times), this logic would just find the first one, right? Did you look to see if there were ever multiple? Also, it's conceivable that one published event could be on one page and another published event could be on another...

inventory.py
pushyamig: Yes, It is probably ok For now. We can have as many packages are we want. If we have to make a package we can make it as `canvas` and have different canvas data under this package like published date, user, enrollment, list of published courses.

published_date/published_date.py
pushyamig: I asked this question before in slack and it was decided that to get the latest published date. And The way the canvas course audit logs work they send the audit events in desc order

inventory.py
ssciolla: I'm just trying to keep this organized. I find it useful to separate imports that are for libraries we maintain in this project and those we don't.

published_date/published_date.py
ssciolla: Okay, I do remember that conversation. Can you please put a comment here that the audit logs are in descending order? Otherwise, this logic isn't clear to me.

published_date/published_date.py
pushyamig: Done

inventory.py
pushyamig: I renamed the package name to the canvas. we can have all the canvas related data in that package

published_date/published_date.py
pushyamig: Yes and I tested if ever the course has more event logs that we can paginate and get more data

inventory.py
ssciolla: This is not a bad idea, but it wasn't what I was asking you to do. I was just asking you to move your import down under the "# local libraries" comment.

ssciolla: Thank you!

published_date/published_date.py
ssciolla: Thanks!

ssciolla:
I think I'm about ready to approve this. @pushyamig, can you add the new environment variables to the table in the README?

pushyamig:
> I think I'm about ready to approve this. @pushyamig, can you add the new environment variables to the table in the README?

Done

Issue #15, opened 2020-03-19
Find a way to test database insertion using only Docker


ssciolla:
Right now, we don't have a way (or at least a documented way) to test either using the application in a Docker container and local MySQL, or with both the MySQL and the application in Docker containers. A `docker-compose.yml` may not be appropriate since this is a job, not a service, but some documentation or shell scripts should explain how to get these scenarios to work.

ssciolla:
After talking about and working on this with @jonespm, I think `docker-compose.yml` will actually work pretty well. I'll put the finishing touches on what I have and open a PR.

PR #19, opened 2020-03-23
Support local development with docker-compose (#15)

ssciolla:
This PR adds a `docker-compose.yml` file, allowing users to run the application from end to end without having a MySQL installation on their machine. Updates were made to the README describing how to configure and use the file. Additionally, a `.dockerignore` file was added to keep data and unnecessary files outside of the images. This PR aims to resolve issue #15. 

Thanks to @jonespm for working with me on this and giving me a crash course on writing `docker-compose.yml` files.


docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+  mysql:
+    image: mysql:5.7
+    restart: on-failure
+    environment:
+      - MYSQL_ROOT_PASSWORD=inventory_root_password
+      - MYSQL_HOST=inventory_mysql
+      - MYSQL_PORT=3306
+      - MYSQL_DATABASE=inventory_local
+      - MYSQL_USER=inventory_test_user
+      - MYSQL_PASSWORD=inventory_test_password
```
pushyamig: Can you make these credentials shorter? 
`MYSQL_ROOT_PASSWORD=ci_root_pw`
`MYSQL_USER=ci_user`
`MYSQL_PASSWORD=ci_test_password`
Call me lazy but we don't want long password for local DB set up.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+    entrypoint: ['docker-entrypoint.sh', '--default-authentication-plugin=mysql_native_password']
+    ports:
+      - "4444:3306"
+    volumes:
+      - ./.data/mysql:/var/lib/mysql:delegated
+    container_name: inventory_mysql
```
pushyamig: a better name would be `course_inventory_mysql`?

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+  job:
+    build:
+        context: .
+        dockerfile: Dockerfile
+    depends_on:
+      - mysql
+    volumes:
+      - ${HOME}/secrets/inventory:/app/config/secrets
+      - ${HOME}/data/inventory:/app/data
+    container_name: inventory
```
pushyamig: can name changed to be course_inventory to sync up with github repo name

docker-compose.yml
ssciolla: Done.

docker-compose.yml
ssciolla: Done.

docker-compose.yml
ssciolla: Done.

ssciolla:
I changed some names in the `docker-compose.yml`, per a request from @pushyamig. Note, if you tried running this before, you will need to rename the directory paths the volumes are mapped to to use `course-inventory`. Also, delete the `.data` directory so the new credential info will be used.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
...
+  mysql:
+    image: mysql:5.7
+    restart: on-failure
+    environment:
+      - MYSQL_ROOT_PASSWORD=ci_root_pw
+      - MYSQL_HOST=course_inventory_mysql
+      - MYSQL_PORT=3306
+      - MYSQL_DATABASE=course_inventory_local
+      - MYSQL_USER=ci_user
+      - MYSQL_PASSWORD=ci_pw
```
pushyamig: I would suggest adding these mysql credentials default values to `env_blank.json` sample as well

ssciolla: I'm gonna hold off on this for now;  not everyone may be using docker-compose. Instruction are provided in the README. Requirements are also changing, we may not shift to using Oracle.

Issue #22, opened 2020-03-25
Include sis_user_id in User table


zqian:
Including sis_user_id in the user table would greatly help the joining process with tables in data warehouse.

zqian:
verified that the SIS_ID is included in the USER table

PR #24, opened 2020-03-25
Add sis_id to user table (#22)

ssciolla:
This PR makes adds a migration and makes modifications to the existing UDW user query in order to add `user.sis_id` (a user's UM ID) to the database. It pulls the `sis_id` from the `sis_user_id` column of the `pseudonym_dim` table. Some processing is done to ensure that the value is an integer. The cases where they are not seem to be for friend accounts, where the value is an email. This PR aims to resolve issue #22.

### DIFF
--- a/inventory.py
+++ b/inventory.py
@@ -144,12 +144,23 @@ def pull_enrollment_data_from_udw(course_ids) -> pd.DataFrame:
     return enrollment_df
 
 
+def process_um_id(id: Union[str, None]) -> Union[int, None]:
+    if id is not None:
+        try:
+            um_id = int(id)
+            return um_id
+        except ValueError:
+            logger.debug(f'Invalid um_id found: {id}')
+    return None
+
+
 def pull_user_data_from_udw(user_ids: Sequence[int]) -> pd.DataFrame:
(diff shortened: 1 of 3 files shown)


inventory.py
```diff
@@ -144,12 +144,22 @@ def pull_enrollment_data_from_udw(course_ids) -> pd.DataFrame:
...
+    return um_id
+
+
 def pull_user_data_from_udw(user_ids: Sequence[int]) -> pd.DataFrame:
     users_string = ','.join([str(user_id) for user_id in user_ids])
     user_query = f'''
         SELECT u.id AS warehouse_id,
                u.canvas_id AS canvas_id,
                u.name AS name,
+               p.sis_user_id AS um_id,
```
pushyamig: Nitpicking sis_id resonates with people more than umid

ssciolla: I'm fine changing it, I was just trying to be direct. Anyone else have opinions?

ssciolla: It's probably better to keep it less UM specific...

zqian: can we use sis_id instead?

ssciolla: Done!

Issue #14, opened 2020-03-18
Add a metadata table to the database schema


ssciolla:
We should add a table to the database (once PR #10 is merged) that provides at least a timestamp of when the data was last updated.

jonespm:
Resolved by #14

lsloan:
:shipit:

Issue #26, opened 2020-03-26
Update README with OpenShift and migration details


ssciolla:
Recent work on this project revealed a couple new or missing details from from the `README`. Specifically, we need to define `USER` as an environment variable in OpenShift for `yoyo-migrations` to work properly. Also, we should add at least a sentence about specifying migration dependencies with `yoyo`.

jonespm:
Resolved by #27

zqian:
Verified README contains details of yoyo-migrations.

PR #27, opened 2020-03-26
Add job_run table, introduce append table concept, update README (#14, #26)

ssciolla:
In order to support data consumers and developers, this PR introduces a `job_run` table to track when the latest job run updating the database was completed. Since this table's records should persist between runs, the concept of an append table was introduced and incorporated into the `DBCreator` class. Lastly, updates to the README were made to reflect the above changes, as well as provide additional guidance on using the migration technology. A task list will be provided below with additional details. The PR aims to resolve issues #14 and #26.

- [x] Add migration for `job_run` table.
- [x] Update `DBCreator` class to include `append_table_names` parameter, adding condition to the `drop_records` method to exclude append tables.
- [x] Create an `APPEND_TABLE_NAMES` environment variable.
- [x] Change use of `DBCreator` in `inventory.py` to use `set_up_database` method.
- [x] Add step to `inventory.py` to insert new `job_run` record with timestamp at the end of the job run.
- [x] Re-order key-value pairs in `env_blank.json` and associated table in the `README`.
- [x] Expand `README` descriptions of a few environment variables.
- [x] Add new `README` entry for `APPEND_TABLE_NAMES`.
- [x] Add details to Openshift Deployment section about `USER` environment variable needed by `yoyo-migrations`.
- [x] Update `README` migration instructions to mention migration dependencies.


ssciolla:
I'm making some changes to the `job_run` tables after getting feedback from @jennlove-um. Shouldn't take too long.

ssciolla:
Okay, this should now be ready for review. The exact structure of the `job_run` table may change over time as we incorporate other jobs or completely move to using APIs.

db/migrations/0005.add_job_run_table.py
```diff
@@ -0,0 +1,18 @@
...
+# file: migrations/0005.add_job_run_table.py
+#
+from yoyo import step
+
+__depends__ = {'0004.add_sis_id'}
+
+step('''
+    CREATE TABLE IF NOT EXISTS job_run
+    (
+        id INTEGER NOT NULL UNIQUE AUTO_INCREMENT,
```
zqian: There will be multiple jobs running to get data from different sources, and those job status will all be stored in this table. 

How about adding a column here for the job_name, e.g. "course_inventory", "Zoom", "MiVideo", etc.

pushyamig:
This PR runs fine and append data new table `job_run`

db/migrations/0005.add_job_run_table.py
ssciolla: @zqian, I can do this, but if so, I do not want to include the `unizin_data_updated_at` here, since some jobs will not require this and I'd rather not have a bunch of null values. How about I create a `data_source_status` table and relate it many to one to `job_run`? Something like this:
Screen Shot 2020-03-27 at 10 07 37 AM

jonespm: Yeah I think this should be more generic. In MyLA we're using the django cron and while that package is unmainted and has a few issues I think their data model is pretty good and does what we need here. We probably don't really two separate tables. Aren't all of the values always going to be populated? 

```
    """
    Keeps track of the cron jobs that ran etc. and any error
    messages if they failed.
    """
    code = models.CharField(max_length=64, db_index=True)
    start_time = models.DateTimeField(db_index=True)
    end_time = models.DateTimeField(db_index=True)
    is_success = models.BooleanField(default=False)
    message = models.TextField(default='', blank=True)  # TODO: db_index=True

    # This field is used to mark jobs executed in exact time.
    # Jobs that run every X minutes, have this field empty.
    ran_at_time = models.TimeField(null=True, blank=True, db_index=True, editable=False)
```

https://github.com/Tivix/django-cron/blob/master/django_cron/models.py

jonespm: I think if we the start/end times in this database it would make it unnecessary to do the logging in the future too. The nice thing about the django cron is to create a job you just extend [CronJobBase](https://github.com/Tivix/django-cron/blob/7fa5c2d2835520d8c920f3c103937ab292d9e545/django_cron/__init__.py#L43) and implement a do() as your method. This could be implemented as part of #33 for a similar pattern to ensure the job_run table is populated as expected.

ssciolla: @jonespm, from my perspective, no, sometimes we won't  have a `data_source_updated_at` value, since we more or less have to assume API data is current. I was trying to handle the possibility of having multiple data sources in a job (which mine currently does).

I don't know how well the CronJob model does this, though some of those fields might be useful.

ssciolla:
I'm closing this since there are some conflicts and I think I want to make these changes as part of my work on issue #33. I'll also make the README updates separately, since it should have been separate any way.

Issue #23, opened 2020-03-25
Reveal sections info for Canvas courses


zqian:
Canvas courses can contain multiple sections. We need to create a separate table to capture this one-to-many relationship. The course_section table should include the following attributes:

1. Canvas course id
2. Section id
3. Section name
4. Section SIS ID
5. Section type: lecture, lab, discussion, etc.

ssciolla:
@zqian, is there a good API endpoint to use for this? I didn't find any way yet to fetch sections in bulk, so I'm concerned that it's going to be similar to the `published_date` thing, where we need to make tens of thousands of requests. I will look into whether GraphQL can surface any of this.

zqian:
@ssciolla try this UDW query:

select * from course_section_dim csd  where course_id like '%<canvas_course_id>'

ssciolla:
Okay, but that's a different approach. I could make this work with UDW fairly quickly, but if we want to move to API eventually...

zqian:
True. Let me search for the API then

zqian:
How about the sections API? https://canvas.instructure.com/doc/api/sections.html#method.sections.index

zqian:
GraphQL?

https://community.canvaslms.com/thread/37311-with-the-canvas-api-how-can-you-get-a-list-of-all-sections-for-a-term

ssciolla:
I will look into GraphQL, though if we do that, we may want to change a bunch of stuff at once.

ssciolla:
Closing this now. Related work continues on this in issue #42.

Issue #28, opened 2020-03-30
Use API queries for enrollment, user, and section data


ssciolla:
Currently, we are using UDW for gathering enrollment and user info. As I look at completing issue #23, we also want to figure out how to efficiently use API calls to get data about these other entities, so that we are looking at more current data in the report. This will likely make the job significantly longer, but hopefully still less than 3 hours.

ssciolla:
Closing this now. Related work continues on this in issue #42.

PR #38, opened 2020-03-31
Collect enrollment, user, and section data using the Canvas API (#23, #28)

ssciolla:
This PR makes significant changes to the Canvas data fetching API process to rely as much as possible on the Canvas API and to incorporate section data. Much of this was accomplished by using a `total_students` attribute to limit subsequent queries to only courses with 1 or more students enrolled and by using asynchronous requests (pioneered by @pushyamig) and GraphQL to gather the data for other entities. To support this, I made a few changes to the MySQL data model (three migrations) to make `canvas_id` the primary key for Canvas data tables, align columns and data types with what is returned from Canvas, and to add a `course_section` table. A task list is provided below with more details on the changes. This PR aims to resolve issues #23 and #28.

- [x] Add `"include": ["total_students"]` to existing Canvas course query using a regular REST endpoint, and only return courses with one or more student enrollments.
- [x] Add a `gql_queries.py` file that contains a GraphQL query fetching enrollments associated with a course, with nested data about the associated user and section.
- [x] Create `AsyncEnrollGatherer` class in `canvas` module that uses the Canvas GraphQL endpoint and `requests-futures`, following and expanding upon the existing `published_date.py` example.
- [x] Remove enrollment fetching from UDW.
- [x] Modify UDW user query to use `canvas_id` as the identifier and only fetch `sis_id` and `uniqname`.
- [x] Add a long migration to drop `warehouse_id` columns from tables and use `canvas_id` as the primary key instead (note that `user.canvas_id` needed to remain a `BIGINT` since occassionally some Canvas IDs are longer). 
- [x] Add a migration modifying the `user` table, dropping the `workflow_state` column (since it wasn't available via the API and wasn't in use by Tableau) and letting `uniqname` be null.
- [x] Add a migration creating a `course_section` table and adding a foreign key constraint to `enrollment` table.
- [x] Remove `WAREHOUSE_INCREMENT` from `inventory.py`, since it's not in use currently (I updated and kept the entry in `README.md`, as it could be useful in the future).
- [x] Remove process of validating `user_ids` in enrollment table, since it is no longer an issue with the Canvas API data. 
- [x] Modify some spacing and formatting in `inventory.py`.

Note: I will be moving `gql_queries.py` and `inventory.py` into the `canvas` module (and I may re-name it) in a future PR to make it easier to run separate jobs.


canvas/async_enroll_gatherer.py
```diff
@@ -0,0 +1,179 @@
...
+        more_to_gather = True
+
+        while more_to_gather:
+            course_ids_to_process = sorted(self.get_course_ids_for_incomplete_enrollments())
+
+            if len(course_ids_to_process) == 0:
+                more_to_gather = False
+                logger.info('Enrollment records have been gathered')
+            else:
+                if (len(prev_course_id_lists) > 2) and (prev_course_id_lists[0] == course_ids_to_process) and (prev_course_id_lists[1] == course_ids_to_process):
```
ssciolla, line 173: This check needs to be improved. I'm to find a clean way to enable re-tries. This MIGHT work some of the time, but is unreliable.

ssciolla: I think I fixed this.

pushyamig:
This PR ran as expected

pushyamig:
I see the DB schema for `course`, `user`,`course_section`, `enrollment` tables use `canvas_id` as `id` in name convention. It was a bit confusing for me and was asking is this canvas_id from other tables or id of this table? I don't know the rationale behind the adaptation. 
Optional: I would suggest id instead of canvas_id.

ssciolla:
@pushyamig , I guess I kept it as `canvas_id` because it's the identifier for that entity in canvas. When I reference a table's `canvas_id` in foreign keys, I do just call `course_id` or `user_id`, depending on what the table is. I can see it being confusing, but I can also seeing `id` being not specific enough. I tend to use `id` for auto-incremented IDs maintained by the database.

canvas/async_enroll_gatherer.py
```diff
@@ -0,0 +1,210 @@
...
+    def make_requests(self, course_ids: Sequence[int]) -> None:
+        with FuturesSession(max_workers=self.num_workers) as session:
+            responses = []
+            for course_id in course_ids:
+                # Prep params
+                params = copy.deepcopy(self.default_params)
+                params['variables']['courseID'] = course_id
+                if course_id in self.course_enrollments.keys():
+                    course_enrollment_dict = self.course_enrollments[course_id]
+                    enroll_page_info = course_enrollment_dict['page_info']
```
pushyamig, line 149: Will the case ever occur when a course already went through the loop and there was some 500 error and put on this retry list again be missed? I am guessing here once a loop is done it will go the `while` routine and further picked up for calling. is this how the logic happening?

canvas/async_enroll_gatherer.py
```diff
@@ -0,0 +1,210 @@
...
+            complete_course_ids = self.get_complete_course_ids()
+            course_ids_to_process = sorted(self.get_incomplete_course_ids())
+
+            if len(course_ids_to_process) == 0:
+                more_to_gather = False
+            else:
+                unstarted_course_ids = [course_id for course_id in course_ids_to_process if course_id not in self.course_enrollments.keys()]
+                # This condition will stop the gather process if all the remaining course_ids have not been started.
+                # Several request attempts will have been made using this course_id before reaching this state.
+                if len(complete_course_ids) > 0 and unstarted_course_ids == course_ids_to_process:
```
pushyamig, line 203: I guess you are checking after the initial round processing if the unfinished list doesn't change the exit out?

canvas/async_enroll_gatherer.py
ssciolla: Yeah, so, if a request is made for a course and it fails (`problem_encountered`) because of an irregular status code or `JSONDecodeError`, then no record will be created in `self.course_enrollments`. When the loop starts over (which will happen once the first page is asked for from every course), it will try this course again. It will keep trying to ask for this course on each loop so long as there are uncompleted courses (courses with more pages). Does that help explain? There may be a better way to do this.

canvas/async_enroll_gatherer.py
ssciolla: Kind of. I'm checking whether at least one course is complete and whether the remaining courses to process are unstarted, i.e. they haven't successfully processed any pages. Until the end, the course IDs should include started but unfinished courses.

db/migrations/0005.change_pks_to_canvas_id.py
```diff
@@ -0,0 +1,79 @@
...
+#
+from yoyo import step
+
+__depends__ = {'0004.add_sis_id'}
+
+steps = [
+    # Remove foreign key constraints
+    step('''
+        ALTER TABLE enrollment
+        DROP FOREIGN KEY enrollment_ibfk_1
```
pushyamig, line 12: What is this `enrollment_ibfk_1`?

db/migrations/0005.change_pks_to_canvas_id.py
```diff
@@ -0,0 +1,79 @@
...
+    '''),
+    step('''
+        ALTER TABLE enrollment
+        DROP FOREIGN KEY enrollment_ibfk_2
```
pushyamig, line 16: and this `enrollment_ibfk_2`

db/migrations/0005.change_pks_to_canvas_id.py
ssciolla: Those are the names of the foreign key constraints. Those were automatically created by the foreign key processes in the first migration. You can see them by running a command, let me find it.

ssciolla: If you just run `SHOW CREATE TABLE enrollment` in `course_inventory_test`, you'll see the constraint names.

ssciolla: `ib` must stand for `InnoDB`, the engine we're using?

db/migrations/0005.change_pks_to_canvas_id.py
```diff
@@ -0,0 +1,79 @@
...
+    step('''
+        ALTER TABLE enrollment
+        ADD CONSTRAINT fk_course_id
+            FOREIGN KEY (course_id)
+            REFERENCES course(canvas_id)
+            ON UPDATE CASCADE ON DELETE CASCADE;
+    '''),
+    step('''
+        ALTER TABLE enrollment
+        ADD CONSTRAINT fk_user_id
```
pushyamig, line 74: This time you being more specific to constraints name so that you can control when you have to change it next time

Issue #40, opened 2020-03-31
Display duplicates removed from zoom script


jonespm:
For some reason the zoom script still has a few duplicates that are removed when we call drop_duplicates [using one of these solutions perhaps.](https://stackoverflow.com/questions/54470620/pandas-how-to-store-rows-dropped-using-drop-duplicates)

jonespm:
This should also just avoid the inplace as mentioned in #52. If @ssciolla wants to pick this up for for it. ;)

jonespm:
Won't fix this one as it's Zoom related

PR #48, opened 2020-04-02
issue_32 bluejeans report

zqian:
please review the draft of bluejeans report

### DIFF
--- a/online_meetings/bluejeans_report.py
+++ b/online_meetings/bluejeans_report.py
@@ -1,0 +1,190 @@
+import requests
+# standard libraries
+import json
+import logging
+import os
+import sys
+from json.decoder import JSONDecodeError
+from typing import Dict, Union
+import yaml
+
+# third-party libraries
+import pandas as pd
+import furl
+import time
+import dateparser
+import pytz
+from datetime import datetime, timedelta, timezone
+import urllib.request
+import glob
+
+# importing required modules 
+from zipfile import ZipFile 
+
+# Initialize settings and globals
+
+logger = logging.getLogger(__name__)
+
+# Default time to sleep when hitting a 429 limit
+DEFAULT_SLEEP_TIME = 10
+
+
+# read configurations
+try:
+    with open(os.path.join(os.path.dirname(__file__), '../config/env.json')) as env_file:
+        ENV = yaml.safe_load(env_file.read())
+except FileNotFoundError:
+    sys.exit(
+        'Configuration file could not be found; please add env.json to the config directory.')
+
+logging.basicConfig(level=ENV.get('LOG_LEVEL', 'DEBUG'))
+
+url = "https://api.bluejeans.com/oauth2/token?Client"
+
+payload = "{ \"grant_type\": \"client_credentials\", \"client_id\": \"Metrics_BJN\", \"client_secret\": \"" + ENV.get('BLUEJEANS_CLIENT_SECRET') + "\"}"
+logger.info(payload)
+headers = {
+  'Accept': 'application/json',
+  'Content-Type': 'application/json'
+}
+
+token_expiration_time = datetime.now()
+response = requests.request("POST", url, headers=headers, data=payload).text.encode('utf8')
+token_json = json.loads(response)
+token = token_json["access_token"]
+logger.info(token)
+## the expiration time in seconds
+expires_in_sec = token_json["expires_in"]
+## minus for one minute, for safty check
+expires_in_sec = expires_in_sec - 60
+token_expiration_time = datetime.now() + timedelta(seconds=expires_in_sec)
+logger.info(f"{token} expires on: {token_expiration_time}")
+
+# construct param string BLUEJEANS_MEETINGS_PARAMS
+
+params = ENV["BLUEJEANS_MEETINGS_PARAMS"]
+f = furl.furl('')
+f.args = params
+logger.info(f.url)
+
+bl_url = ENV["BLUEJEANS_MEETINGS_URL"]
+url = f'{bl_url}{f.url}'
+logger.info(url)
+
+payload = {}
+headers = {
+  'Accept': 'application/json',
+  'Content-Type': 'application/json',
+  'Authorization': f'Bearer {token}'
+}
+
+page_size = 1000
+total_records = 0
+
+total_df = pd.DataFrame()
+
+params = ENV["BLUEJEANS_MEETINGS_PARAMS"]
+
+# TODO: the earliest time is hardcoded now
+early_date = dateparser.parse(ENV.get('ZOOM_EARLIEST_FROM', '2020-03-31')).date()
+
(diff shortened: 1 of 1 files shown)


online_meetings/bluejeans_report.py
```diff
@@ -0,0 +1,186 @@
...
+
+# importing required modules
+from zipfile import ZipFile
+
+# Initialize settings and globals
+
+logger = logging.getLogger(__name__)
+
+# Default time to sleep when hitting a 429 limit
+DEFAULT_SLEEP_TIME = 10
```
jonespm, line 26: This should come from the env, move it down after the env is read
`DEFAULT_SLEEP_TIME = ENV.get('DEFAULT_SLEEP_TIME', 10)`

online_meetings/bluejeans_report.py
```diff
@@ -0,0 +1,186 @@
...
+
+# read configurations
+try:
+    with open(os.path.join(os.path.dirname(__file__), '../config/env.json')) as env_file:
```
jonespm, line 30: Sam said the path for this is ../config/secrets/env.json.

This will be updated in a central manager but we might as well just put all our files in the same place.

online_meetings/bluejeans_report.py
```diff
@@ -0,0 +1,186 @@
...
+        ENV = yaml.safe_load(env_file.read())
+except FileNotFoundError:
+    sys.exit(
+        'Configuration file could not be found; please add env.json to the config directory.')
+
+logging.basicConfig(level=ENV.get('LOG_LEVEL', 'DEBUG'))
+
+url = "https://api.bluejeans.com/oauth2/token?Client"
+
+payload = "{ \"grant_type\": \"client_credentials\", \"client_id\": \"Metrics_BJN\", \"client_secret\": \"" + ENV.get('BLUEJEANS_CLIENT_SECRET') + "\"}"
```
jonespm, line 40: You should be able to just define this as a dictionary rather than a string, would save on some quote escaping. If it's an API that needs json encoding the you should still define it as a dict and when posting it call `data=json.dumps(payload)`. Requests also lets you call `json=payload`.

online_meetings/bluejeans_report.py
```diff
@@ -0,0 +1,186 @@
...
+## the expiration time in seconds
+expires_in_sec = token_json["expires_in"]
+## minus for one minute, for safty check
+expires_in_sec = expires_in_sec - 60
+token_expiration_time = datetime.now() + timedelta(seconds=expires_in_sec)
+logger.info(f"{token} expires on: {token_expiration_time}")
+
+# construct param string BLUEJEANS_MEETINGS_PARAMS
+
+params = ENV["BLUEJEANS_MEETINGS_PARAMS"]
```
jonespm, line 61: I don't see what you're using either of these variables `f` or `url` for. I removed furl from the other method as it wasn't used.

online_meetings/bluejeans_report.py
```diff
@@ -0,0 +1,186 @@
...
+for i in range((datetime.now(pytz.timezone('America/Detroit')).date() - early_date).days):
+    param_date = early_date + timedelta(days=i)
+    start = datetime(param_date.year, param_date.month, param_date.day, 0, 0, 0)
+    end = datetime(param_date.year, param_date.month, param_date.day, 23, 59, 59)
+    logger.info("start")
+    logger.info(start.astimezone().isoformat())
+    logger.info("end")
+    logger.info(end.astimezone().isoformat())
+    
+    url = "https://indigo-api.bluejeans.com/v1/enterprise/" + str(ENV.get('BLUEJEANS_ENTERPRISE_ID')) + "/indigo/reportJob/meetings?filter=\
```
jonespm, line 98: You don't have to construct the URL like this, if you have the params as a dict you can pass it with the params argument to requests.request. I think this would probably be more readable if you were using [str.format](https://realpython.com/python-string-formatting/#2-new-style-string-formatting-strformat) if you didn't want to do this.

online_meetings/bluejeans_report.py
```diff
@@ -0,0 +1,186 @@
...
+    payload = {}
+    headers = {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json',
+        'Authorization': f'Bearer {token}'
+    }
+    r = requests.get(download_url, headers=headers)
+
+    # specifying the zip file name
+    zip_file_name = "./bj.zip"
```
jonespm, line 154: This won't work in the container. We're probably going to have to create a storage volume, pass in a path via ENV and use that to write. (Like /tmp for instance)

online_meetings/bluejeans_report.py
```diff
@@ -0,0 +1,186 @@
...
+        f.write(r.content)
+    
+    # opening the zip file in READ mode
+    with ZipFile(zip_file_name, 'r') as zip_file:
+        # printing all the contents of the zip file
+        zip_file.printdir()
+    
+        # extracting all the files
+        print('Extracting all the files now...')
+        zip_file.extractall()
```
jonespm, line 166: This is going to have to take the same path as where the file is written.

online_meetings/bluejeans_report.py
```diff
@@ -0,0 +1,186 @@
...
+        for csv_file in glob.glob("./meetings_*.csv"):
+            logger.info(csv_file)
+            df = pd.read_csv(csv_file)
+
+            # append to the total output
+            total_df = total_df.append(df)
+            logger.info(df)
+
+    # remove zip file
+    os.remove(zip_file_name)
```
jonespm, line 176: We probably should also clean up the files that were extracted

online_meetings/bluejeans_report.py
```diff
@@ -0,0 +1,186 @@
...
+    print(f" {zip_file_name} file Removed!")
+    
+
+
+# Remove any duplicate uuids in the record
+logger.info(f"Initial dataframe size: {len(total_df)}")
+if "meetingUUID" in total_df:
+    total_df.drop_duplicates("meetingUUID", inplace=True)
```
jonespm, line 184: I can change this on #40 when I fix this for zoom

jonespm:
I left a number of comments, feel free to click the 3 dots and reference in new issue if you'd like me or someone else to address those in a future fix.

online_meetings/bluejeans_report.py
jonespm: New issue as #55

online_meetings/bluejeans_report.py
jonespm: New issue as #55

online_meetings/bluejeans_report.py
jonespm: Created as new issue

online_meetings/bluejeans_report.py
jonespm: Created as new issue

PR #50, opened 2020-04-02
Fixes #43 - Implements paging for Zoom and Refactors request/retry

### LINKED ISSUE #43: Fix paging and retry/request in Zoom script
- [x] The zoom script currently only retrieves the first page, we should add paging to this script.
- [x]  Also the request/retry loop should be refactored to catch all requests

jonespm:


online_meetings/report.py
```diff
@@ -36,27 +37,50 @@
...
-# Functions
-def get_total_page_count(url: str, headers: Dict[str, Union[str, int]] = None, params: Dict[str, Union[str, int]] = None):
-    if params is None:
-        params = {}
-    if headers is None:
-        headers = {}
+def get_request_retry(url: str, headers: Dict[str, Union[str, int]],
+                      params: Dict[str, Union[str, int]]) -> requests.Response:
+
+    response = requests.request("GET", url, headers=headers, params=params)
```
ssciolla, line 43: I'm trying to run this, but it seems like the `url` is only the URL ending. Am I missing something?

ssciolla: Sorry, zoom base url is not set in the config.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -52,6 +52,9 @@ def zoom_course_report(canvas_account=1, enrollment_term_id=1, published=True):
     per_page = 100
     # Get all published courses from the defined enrollment term
     courses = account.get_courses(enrollment_term_id=enrollment_term_id, published=published, per_page=per_page)
+    # For testing
+    # course = CANVAS.get_course(331376)
+    # courses = [course, ]
```
ssciolla, line 57: Do we need this?

online_meetings/canvas_zoom_meetings.py
```diff
@@ -97,26 +100,36 @@ def zoom_course_report(canvas_account=1, enrollment_term_id=1, published=True):
...
-                            'topic': meeting['topic'],
-                            'join_url': meeting['joinUrl'],
-                            'start_time': meeting['startTime'],
-                            'status': meeting['status'],
-                            'timezone': meeting['timezone']
-                        })
+
+                    # Page iterator
+                    page_num = 1
+                    while True:
```
ssciolla, line 106: Nitpick: don't love `while True ... break`.

online_meetings/canvas_zoom_meetings.py
jonespm: No, I was thinking about removing it, but when I need to test I forget what the call is. Maybe I need to just make it more configurable really, either run on a term or on a specific course.

online_meetings/canvas_zoom_meetings.py
jonespm: Yeah, I don't either. I was thinking about refactoring this but it would be a lot more code. This API call isn't great and you have to make one call first. Then it tells you the number of items per page and the total items. You could set up a for loop based on that information or just break when it runs out of pages to return. 

I don't like while True because you might get into an endless loop. I'm also using a 
`while response.status_code == requests.codes.too_many_requests:` which is essentially a potential endless loop too if it's always returning too_many_requests. This shouldn't be as likely though. I think I left this in from the quick hack but maybe this can be refactored to be better.

ssciolla: So far, it looks like these loops are okay. I don't think it would take too much work to do a simple refactor where you define a variable holding a boolean before. Like what I'm doing in `inventory.py`:

https://github.com/tl-its-umich-edu/course-inventory/blob/373ea575e5ea6830f68a022a02cd0d684a758002/inventory.py#L99-L122

online_meetings/canvas_zoom_meetings.py
ssciolla: Perhaps, yeah. I like including a test mode in some of my projects, which just slices lists down to a reasonable number before processing. I haven't gotten around to implementing that here yet.

online_meetings/report.py
```diff
@@ -19,7 +19,8 @@
 
 # read configurations
 try:
-    with open(os.path.join(os.path.dirname(__file__), '../config/env.json')) as env_file:        ENV = yaml.safe_load(env_file.read())
+    with open(os.path.join(os.path.dirname(__file__), '../config/env.json')) as env_file:
```
ssciolla, line 22: Don't know if you want to do this as part of this PR, but it would make sense to use `config/secrets/env.json` instead, since that's where `inventory` expects it to be. I missed that at first and couldn't figure out for a while none of the authentication was working.

online_meetings/report.py
```diff
@@ -36,27 +37,50 @@
...
+                f"No Retry-After header, setting sleep loop for {DEFAULT_SLEEP_TIME} seconds")
+        while response.status_code == requests.codes.too_many_requests:
+            time.sleep(sleep_time)
+            response = requests.request(
+                "GET", url, headers=headers, params=params)
+
+    # If it's not okay at this point, raise an error
+    if response.status_code != requests.codes.ok:
+        logger.warning(f'Received irregular status code: {response.status_code}')
+        response.raise_for_status()
```
ssciolla, line 65: May not need this if you're going to do the same thing just outside of this function.

ssciolla: Wait, yes, you do. Maybe the logging warning could go away, since you seem to be logging in the `try/except` clauses that wrap uses of this function.

online_meetings/report.py
```diff
@@ -36,27 +37,50 @@
...
+    try:
+        response = get_request_retry(url, headers, params)
+    except requests.exceptions.HTTPError:
+        logger.exception('Received irregular status code during request')
+        return 0
+    try:
+        results = json.loads(response.text.encode('utf8'))
+        total_page_count = results['page_count']
+    except json.JSONDecodeError:
+        logger.warning('JSONDecodeError encountered')
```
ssciolla, line 82: Not sure what weird JSON responses happen with this API, but with Canvas, sometimes the response just is invalid. I  think re-trying in those instances could make sense, so maybe at some point these `try/except`s could get moved into `get_request_retry`?

online_meetings/canvas_zoom_meetings.py
jonespm: I'll create a new issue that will reference removing this and supplying a term or course as part of a refactor. It's on #53. We'll do this after.

online_meetings/report.py
jonespm: Oh I didn't notice that, I think this will be refactored completely anyway when the job runner gets committed but I can make this change and move it to secrets.

online_meetings/report.py
jonespm: Yeah sounds good, I had it in there before I refactored it to throw an exception. This seemed to be the preferred pattern for handling these response errors.

online_meetings/report.py
jonespm: Yeah, this `get_request_retry` doesn't catch the Canvas one yet. get_total_page_count is inside of a get_request_retry pattern though which is mostly written to catch the zoom issue of REALLY short throttling. I'd prefer everything with canvas go through the canvasapi package eventually.

jonespm: I'm just going to mark this one as resolved for now, we probably need better API error handling in a central place.

online_meetings/canvas_zoom_meetings.py
jonespm: I think I'll just improve it to only iterate through the expected number of pages. This was really just the quickie hack.

ssciolla: Yeah, if you can change to `for` loop, that's probably better all around.

jonespm: Okay I rewrote this entire file for this ;)

jonespm:
I created a new issue to fix the docstrings in a future PR. I think we should have docstrings for all of our methods.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -43,95 +45,117 @@
...
+        r = self.zoom_session.post(url=url, data=data)
+        # Get the XSRF Token
+        pattern = re.search('"X-XSRF-TOKEN".* value:"(.*)"', r.text)
+        if pattern:
+            self.zoom_session.headers.update({
+                'X-XSRF-TOKEN': pattern.group(1)
+            })
+
+            # Page iterator
+            page_num = 1
```
ssciolla, line 82: I think this `page_num` can get deleted right, since you're using a loop var?

jonespm: Yup thanks that one wasn't used, thought I caught them all!

jonespm:
I ended up redoing this this into a class because 

1. Everything was updating the same zoom_courses and zoom_courses meeting list and it seemed silly passing them around
2. The zoom session needed to be shared among all these classes
3. This will eventually be a more generic LTIPlacements and probably subclassed to Zoom/BlueJeans/etc so figure I should just get this going now
4. I needed to reduce the complexity and break some pieces out (like retrieving pages from zoom so we could get the first page) and it just seemed to be easier to continue breaking this up

There are still some things incomplete in here for sure that I intend to fix in the future and already have issues for.

PR #60, opened 2020-04-03
Fixes #53 - Refactor zoom classes to supply a term or course

### LINKED ISSUE #53: Refactor zoom classes to supply a term or course
When running the "canvas_zoom_meetings" it should either allow to supply an entire term to report on or a list of individual courses. This should also remove the test code that is present around like 55 for testing an individual course.

_Originally posted by @ssciolla in https://github.com/tl-its-umich-edu/course-inventory/pull/50_

jonespm:


README.md
```diff
@@ -51,7 +51,8 @@ To configure the application before installation and usage (see the next section
     ----- | -----
     `LOG_LEVEL` | The minimum level for log messages that will appear in output. `INFO` or `DEBUG` is recommended for most use cases; see [Python's logging module](https://docs.python.org/3/library/logging.html).
     `CANVAS_ACCOUNT_ID` | The Canvas instance root account ID number associated with the courses for which data will be collected.
-    `CANVAS_TERM_ID` | The Canvas instance term ID number that will be used to limit the query for Canvas courses.
+    `CANVAS_TERM_ID` | The Canvas instance term ID number that will be used to limit the query for Canvas courses. Set to 0 to just use ADD_COURSE_IDS.
+    `ADD_COURSE_IDS` | Additional CourseID's to retrieve. Duplicates found in CANVAS_TERM_ID (if defined) are not currently removed.
```
ssciolla, line 55: It looks a little awkward, but could we figure out how to drop duplicates? Maybe some kind of mapping function that checks the course ID on the course object? Seems like something good to do in preparation for database insertion, though you could just drop duplicates later. It'll be a couple less requests though! 

I think it might streamline the code a little if you did that first, and then iterated over the unique course objects all at once.

ssciolla: Also feeling like this could be incorporated in other parts of the codebase, but that can definitely wait.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -105,51 +105,65 @@ def get_zoom_details(self, url: str, data: Dict[str, str], course_id: int):
...
         per_page = 100
         # Get all published courses from the defined enrollment term
-        courses = account.get_courses(enrollment_term_id=enrollment_term_id, published=published, per_page=per_page)
-        # For testing
+        courses = []
+        if enrollment_term_id:
+            courses = account.get_courses(enrollment_term_id=enrollment_term_id, published=published, per_page=per_page)
+
+        # If there's course_id's passed in extend this to include these courses
+
```
ssciolla, line 148: Maybe this comment can be moved to be with wherever you're dealing with the course IDs?

README.md
```diff
@@ -51,7 +51,8 @@ To configure the application before installation and usage (see the next section
     ----- | -----
     `LOG_LEVEL` | The minimum level for log messages that will appear in output. `INFO` or `DEBUG` is recommended for most use cases; see [Python's logging module](https://docs.python.org/3/library/logging.html).
     `CANVAS_ACCOUNT_ID` | The Canvas instance root account ID number associated with the courses for which data will be collected.
-    `CANVAS_TERM_ID` | The Canvas instance term ID number that will be used to limit the query for Canvas courses.
+    `CANVAS_TERM_ID` | The Canvas instance term ID number that will be used to limit the query for Canvas courses. Set to 0 to just use ADD_COURSE_IDS.
```
ssciolla, line 54: I realized the `0` is going to be a problem here if you want to run your jobs and `inventory.py`, since I treat `CANVAS_TERM_ID` as required. How should we deal with that? I could probably change the `inventory` code a bit.

ssciolla: If it's only used for testing, it might not be a problem.

README.md
jonespm: Yeah, I think we *could* do it. We'd either have to skip those courses defined when going through the term for loop or keep track of these and not process them afterward. I'm feeling like real world usage isn't that you'll specify courses and a term, that it would be one or another. There's no reason to add courses that are part of the term as it will already be processed. We can see if this is really an issue. :)

jonespm: And yeah, I'd like to refactor this code (again) to use the threading calls too so that may end up making this a lot more straightforward to do.

README.md
jonespm: I feel like we'll pretty much be doing full term runs with this and this is mostly being added for testing purposes.

ssciolla: Sure, that make sense

online_meetings/canvas_zoom_meetings.py
jonespm: Thanks, I originally had it going first but was thinking of adding in a process to catch dupes. I guess I can just add an extra if statement to remove dupes. Duh it's really easy.

jonespm:
Just hitting request re-review because I am removing dupes now.

README.md
```diff
@@ -51,7 +51,8 @@ To configure the application before installation and usage (see the next section
...
+    `ADD_COURSE_IDS` | Additional CourseID's to retrieve. Duplicates found in CANVAS_TERM_ID (if defined) will be removed.
```
ssciolla, line 55:
```suggestion
    `ADD_COURSE_IDS` | Additional Canvas course ID's to retrieve. Duplicates found in CANVAS_TERM_ID (if defined) will be removed.
```

online_meetings/canvas_zoom_meetings.py
```diff
@@ -105,51 +105,66 @@ def get_zoom_details(self, url: str, data: Dict[str, str], course_id: int):
...
-        # For testing
+        courses = []
+        if enrollment_term_id:
+            courses = account.get_courses(enrollment_term_id=enrollment_term_id, published=published, per_page=per_page)
+
         course_count = 0
         for course in courses:
             course_count += 1
+            if add_course_ids and course.id in add_course_ids:
+                add_course_ids.remove(course.id)
```
ssciolla, line 151: This looks okay and should work. You could also create a list of all the course IDs from the object, and use a different condition in the later for loop. Same result, but you wouldn't have to remove and re-add, which is a tiny bit silly : )

online_meetings/canvas_zoom_meetings.py
```diff
@@ -105,51 +105,66 @@ def get_zoom_details(self, url: str, data: Dict[str, str], course_id: int):
...
-                    self.zoom_courses.append({'account_id': course.account_id, 'course_id': course.id, 'course_name': course.name})
-
-                    fields = form.findAll('input')
-                    formdata = dict((field.get('name'), field.get('value')) for field in fields)
-                    # Get the URL to post back to
-                    posturl = form.get('action')
-                    self.get_zoom_details(posturl, formdata, course.id)
+            self.get_zoom_course(course)
+
+        # If there's course_id's passed in extend this to include these courses
```
ssciolla, line 157:
```suggestion
        # If there are course_ids passed in, also process those
```

online_meetings/canvas_zoom_meetings.py
ssciolla: Creating the list of IDs from the objects might be more or less resource-intensive than requesting the same course twice, not sure.

jonespm: Yeah, it was either
1. Check and remove as it's processing
2. Add it to a long list and compare/remove afterward. 

This performance is going to have very negligible impact on the api call. Though I tested it an in a simulation of 1000 "courses" iterating over 300000 "courses" the solution that removes as it's going along is about 30% faster than the one that creates a new list. 

Though overall the fastest process took added about 3 seconds. For only 10 courses it's in the milliseconds.

jonespm: Interesting though. But one api call is going to be slower than this entire process unless the defined course list was well over 1000.  (Which is not going to happen)

ssciolla: Yeah, I think you would have to make the same call twice for each added course ID, given the current configuration.

PR #61, opened 2020-04-03
Fix the member variables treating as tuple

pushyamig:
Problem: member variable string gets treated as Tuple in Python
Since I declared then as such. This PR should fix this case.

This is not a bug but minor annoyance
https://stackoverflow.com/questions/27654671/member-variable-string-gets-treated-as-tuple-in-python


pushyamig:
I did not pay much attention. It's just a mistake

Issue #33, opened 2020-03-30
Write a central manager script for this project


jonespm:
This project is becoming a combination of a number of reporting scripts. We should (for now) have a central manager that will start the appropriate report.

- [x] Refector the reports into packages/classes so they can be launched by a central script
- [x] Create a script at the top level like run_reports.py. This should take a command line option for which report(s) to run. 
- [x] This central script should take care of things like report timing, config loading, exception handling and anything else that might be common to each job.
- [x] After loading env file use values from the environment if they are present
I think it would just be something like this `ENV.update((key, os.environ[key]) for key in ENV.keys())`

ssciolla:
@jonespm, I started working on this. I can go down the command-line argument path, but I was wondering whether we might ever want to run multiple jobs. What if we defined the names in the `env` file? Would that be bad for some reason?

jonespm:
I was just expecting we might have multiple jobs in Jenkins like
Run Zoom Reports
Run Canvas Reports
Run Blue Jeans Reports

We could have each one use a different env file I guess, or pass different command line args. It seems like both could work the same way, so it doesn't matter how you want to implement it. I also have an idea for overriding the values in the ENV file with local environment values that I was going to do for this and MyLA so we could just do that too and keep it simpler. 

So the Jenkins job could like define local env of
REPORTS_TO_RUN=zoom,bluejeans
python run_reports.py

And the REPORTS_TO_RUN from os.environ would override the value(s) in the file.

ssciolla:
Okay, how about something like this, @jonespm? 

https://github.com/tl-its-umich-edu/course-inventory/compare/master...ssciolla:issue-33-create-job-manager

jonespm:
Yes, that is looking pretty good Sam. I just read through it but it's pretty much what I was thinking. Here's some things we'll have to think about.

I think each "module" should be responsible for its own migrations and this central job should run them all (kind of like how Django works). The Zoom tables aren't dependent on the Canvas tables for example. So we might need some changes to the DBCreator. (Haven't looked into that)

Seems like a good start though!

Issue #57, opened 2020-04-03
Add documentation (docstrings) to all methods


jonespm:
All methods should include documentation about what they do. Most of this code doesn't have that.

We've been using the Sphinx format for other projects
https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html

This will fix some Codacy issues related to docstrings and also make the code easier to maintain in the future.

PR #62, opened 2020-04-06
Create job manager, enable append tables, add metadata tables (#14, #33)

ssciolla:
This PR aims to enable multiple jobs (beyond the first, `inventory.py`) by creating a `run_jobs.py` as the Docker start point. Loading and management of environment variables has also been extracted to its own file (`environ.py`) so it could be loaded as needed by other parts of the application. `run_jobs.py` also records job history by updating two new database tables, `job_run` and `data_source_status`. Records should not be dropped from these tables, so I also introduced an "append table" concept, allowing some tables to be preserved by specifying them in `env.json`. Some additional details are provided below in the task list. This PR aims to resolve issues #14 and #33.

- [x] Move `inventory.py` and `gql_queries.py` into `canvas` and rename to `course_inventory`.
- [x] Enable what I'm calling "Append Tables" by updating `db/db_creator.py`.
- [x] Extract `env.json` loading from various places and create `environ.py`
- [x] Implement `Job` and `JobManager` classes in a new `run_jobs.py` file
- [x] Create `job_run` and `data_source_status` tables for storing job metadata
- [x] Add metadata validation and creation to `run_jobs.py`, including creation of `ValidJobName` and `ValidDataSourceName` enums.
- [x] Update `course_inventory.py/inventory.py` to use `APPEND_TABLE_NAMES` and `environ.py`, and to return data source dictionaries that `job_run` uses to create `data_source_status` database records.
- [x] Update README with a generalized overview, new variables, and a section describing how to implement a new job.


ssciolla:
I've opened this for review. I will work on adding a section to the README providing instructions on implementing a new job.

run_jobs.py
```diff
@@ -0,0 +1,135 @@
...
+            job.run()
+
+
+if __name__ == '__main__':
+    how_started = os.environ.get('HOW_STARTED', None)
+
+    if how_started == 'DOCKER_COMPOSE':
+        logger.info('Waiting for the MySQL turtle, hehe')
+        # Wait for MySQL container to finish setting up
+        time.sleep(30.0)
```
ssciolla, line 127: @jonespm, I had to add this because on the first startup (when there is no `.data` directory), it takes a while to finish setting up and the subsequent `db_creator_obj.migrate()` call was failing. Is there a way to specify in `docker-compose` that the job container needs to wait for the MySQL one? This was just a hack to get things working. Also, it's annoying because the first time you need about 30 seconds, but if the `.data` directory exists, it takes about 3 seconds.

jonespm: I'm doing this in the MyLA start script with the nc command which should be installed by default. 
```
echo "Waiting for DB"
while ! nc -z ${MYSQL_HOST} ${MYSQL_PORT}; do   
  sleep 1 # wait 1 second before check again
done
```

I believe the we could have a shell script before this python or could just do this similar check in python with nc. (I didn't test this code below)

```
while not os.system(f"nc -z {os.getenv("MYSQL_HOST")} {os.getenv("MYSQL_PORT")}):
    time.sleep (1)
```

ssciolla: Okay, I'll work on this now.

ssciolla: Kinda remembered something like this, forgot to check MyLA

environ.py
```diff
@@ -0,0 +1,29 @@
...
+
+logging.basicConfig(level=ENV['LOG_LEVEL'])
+
+# Add ENV key-value pairs to environment, skipping if the key is already set
+logger.debug(os.environ)
+for key, value in ENV.items():
+    if key in os.environ:
+        ENV[key] = value
+        logger.info('ENV value overidden')
+        logger.info(f'key: {key}; value: {value}')
```
ssciolla, line 29: This can't handle collection objects. Do you think I should have a JSON load step in here?

ssciolla: Wait, also there's a bug, blah

create_db.py
```diff
@@ -1,30 +1,21 @@
...
 
-try:
-    config_path = os.getenv("ENV_PATH", os.path.join('config', 'secrets', 'env.json'))
-    with open(config_path) as env_file:
-        ENV = json.loads(env_file.read())
-except FileNotFoundError:
-    logger.error('Configuration file could not be found; please add env.json to the config directory.')
-
 DB_PARAMS = ENV['INVENTORY_DB']
+APPEND_TABLE_NAMES = ENV.get('APPEND_TABLE_NAMES', ['job_run'])
```
jonespm, line 13: I feel like we should have DEFAULT_APPEND_TABLES that we don't have to define in the config. Are we ever going to want these dropped automatically by the script?

like 
DEFAULT_APPEND_TABLES = ['job_run', "data_source_status"]
APPEND_TABLE_NAMES = ENV.get('APPEND_TABLE_NAMES', []).extend(DEFAULT_APPEND_TABLES)

environ.py
```diff
@@ -0,0 +1,29 @@
...
+        ENV = json.loads(env_file.read())
+except FileNotFoundError:
+    logger.error('Configuration file could not be found; please add env.json to the config directory.')
+    ENV = None
+
+logging.basicConfig(level=ENV['LOG_LEVEL'])
+
+# Add ENV key-value pairs to environment, skipping if the key is already set
+logger.debug(os.environ)
+for key, value in ENV.items():
```
jonespm, line 25: Cool, I think this is the fix for MyLA https://github.com/tl-its-umich-edu/my-learning-analytics/issues/787

environ.py
ssciolla: Okay, this should be fixed now. I'll do some testing.

jonespm: Hmm, do you mean configure environment variables as JSON and load them? We could [try/except with a json.loads](https://www.geeksforgeeks.org/python-check-whether-a-string-is-valid-json-or-not/) and if it excepts just set the ENV[key] = value otherwise set it to the json value? 

Like if we had an environment variable.
`UDW = {"host": "", "port": "","dbname": "","user": "", "password": ""}`

This would load up the json, but if we had 
CANVAS_TERM_ID = 164 that would also work.

ssciolla: Just did that, as well as fixed the bug I mentioned.
See commit https://github.com/tl-its-umich-edu/course-inventory/pull/62/commits/eadda6888773eb69085425b02e47e6a6a9088a79

create_db.py
ssciolla: I created an issue, #66.

README.md
```diff
@@ -3,12 +3,10 @@
 
 ## Overview
 
-The course-inventory application is designed to gather current-term Canvas LMS data about courses, enrollments, and users in order to inform leadership at the University of Michigan about the status and utilization of Canvas courses. Currently, the application collects data from the Canvas API and the Unizin Data Warehouse and then stores the data in an external MySQL database. Tableau dashboards and other processes then consume that data to generate reports and visualizations.
-
+The course-inventory application is designed to gather current-term Canvas LMS data about courses, enrollments, users, and course activity -- as well as data about the usage of technology for remote learning, including BlueJeans, Zoom, and MiVideo -- in order to inform leadership at the University of Michigan about the status and utilization of tools for teaching and learning. Currently, the application collects data from various APIs and data services managed by Unizin Consortium and then stores the data in an external MySQL database. Tableau dashboards and other processes then consume that data to generate reports and visualizations.
```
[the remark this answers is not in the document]
ssciolla: That's a good point. In general, I just avoid using characters when they aren't technically necessary, but I guess you are making the point that they are sometimes. I will create an issue here.

run_jobs.py
ssciolla: I created a new issue, #67.

README.md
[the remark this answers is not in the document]
ssciolla: I created an issue, #68.

course_inventory/inventory.py
```diff
@@ -169,9 +160,9 @@ def process_sis_id(id: str) -> Union[int, None]:
     return sis_id
 
 
-def run_course_inventory() -> None:
+def run_course_inventory() -> Sequence[Dict[str, Union[str, pd.Timestamp]]]:
```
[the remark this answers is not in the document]
ssciolla: Yeah, I'm not the habit yet of creating these. We have an issue for this that Matt created, #57. We'll address this separately.

course_inventory/inventory.py
```diff
@@ -214,11 +205,30 @@ def run_course_inventory() -> None:
...
     # Pull SIS user data from Unizin Data Warehouse
     udw_user_ids = user_df['canvas_id'].to_list()
-    sis_user_df = pull_sis_user_data_from_udw(udw_user_ids)
+    sis_user_df = pull_sis_user_data_from_udw(udw_user_ids, udw_conn)
     user_df = pd.merge(user_df, sis_user_df, on='canvas_id', how='left')
 
+    # Record data source info for UDW
+    udw_meta_df = pd.read_sql('SELECT * FROM unizin_metadata;', udw_conn)
+    udw_update_datetime_str = udw_meta_df.iloc[1, 1]
+    udw_update_datetime = pd.to_datetime(udw_update_datetime_str, format='%Y-%m-%d %H:%M:%S.%f%z')
```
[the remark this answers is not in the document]
ssciolla: Not quite sure what you're suggesting regarding the timezone. I feel like the UDW `canvasdatadate` is a little bizarre, but it does seem close to the ISO standard. This was needed though I think by `pd.to_datetime`.

db/migrations/0009.add_meta_tables.py
```diff
@@ -0,0 +1,33 @@
...
+from yoyo import step
+
+__depends__ = {'0008.canvas_usage_table'}
+
+step('''
+    CREATE TABLE IF NOT EXISTS job_run
+    (
+        id INTEGER NOT NULL UNIQUE AUTO_INCREMENT,
+        job_name VARCHAR(50) NOT NULL,
+        started_at DATETIME NOT NULL,
```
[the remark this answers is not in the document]
ssciolla: @pushyamig went with `DATETIME` earlier, so I've just gone with it. I feel like maybe there could be less confusion if we just always write UTC timestamps to the database, rather than allowing MySQL to handle that part of the process. Let's discuss later.

course_inventory/inventory.py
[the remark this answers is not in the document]
lsloan: I just meant that `%z`, which is replaced with the currently active time zone, could be replaced with `+00` or simply `Z`, which represents UTC.

PR #64, opened 2020-04-06
Update base URL for UM API Directory in env_blank.json

### LINKED ISSUE #63: env_blank.json includes wrong base URL for UM API Directory
Currently, `env_blank.json` includes the base URL for the UM API Directory as `https://apigw.it.umich.edu` when it should be `https://apigw.it.umich.edu/um`. This needs to be corrected to minimize confusion for new contributing developers.

ssciolla:
This PR adds `/um` to the end of the UM API Directory, which is needed to successfully make requests using the `umich_api` library. The PR aims to resolve issue #63.

Resolves #63.


lsloan:
I committed this change to my `65-mivideo` branch in my fork:

* https://github.com/lsloan/course-inventory/commit/c2ca60342e932c499dde52139614a3ba06ee3d81

jonespm:
I don't really think we should have this specific version in the env_blank here, though this new URL is correct.

ssciolla:
@jonespm, can you elaborate on why you don't think this should be provided here? Is there a security concern, or do you just want to not assume UM is using this? The API Directory URL is weird enough that I think this is useful to contributors, unless there is some concern I'm not aware of.

jonespm:
I guess it's fine, I didn't think this URL was super public but it is coming up in some Google results.

config/env_blank.json
```diff
@@ -3,8 +3,9 @@
   "JOB_NAMES": ["COURSE_INVENTORY"],
   "CANVAS_ACCOUNT_ID": 1,
   "CANVAS_TERM_ID": 164,
+  "API_BASE_URL": "https://apigw.it.umich.edu/um",
```
jonespm, line 6:
```suggestion
```

jonespm:
I think I screwed up this simple change trying to resolve conflicts from Github UI :/

lsloan:
@jonespm, although you commented, no approval was given.  I'm not sure whether you approve or disapprove of this change from what you wrote, so I tagged you for review again.

lsloan:
Who's assigned to merge this PR?

PR #70, opened 2020-04-07
#69 adding FK relationship for canvas_course_usage

### LINKED ISSUE #69: Adding a Foreign key relationship for course_id in canvas_course_usage table
Establish an FK relationship for canvas_course_usage with course table

pushyamig:
Fixes #69

### DIFF
--- a/db/migrations/00010.add_fkc_course_usage_table.py
+++ b/db/migrations/00010.add_fkc_course_usage_table.py
@@ -1,9 +1,9 @@
 #
-# file: migrations/0009.add_fkc_course_usage_table.py
+# file: migrations/00010.add_fkc_course_usage_table.py
 #
 from yoyo import step
 
-__depends__ = {'0008.canvas_usage_table'}
+__depends__ = {'0009.add_meta_tables'}
 
 step('''
         ALTER TABLE canvas_course_usage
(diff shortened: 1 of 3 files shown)


db/migrations/00010.add_fkc_course_usage_table.py
```diff
@@ -0,0 +1,14 @@
+#
+# file: migrations/00010.add_fkc_course_usage_table.py
```
ssciolla, line 2: Can you fix the numbering here and in the file name so it's `0010`?

pushyamig: Done

db/migrations/0010.add_fkc_course_usage_table.py
```diff
@@ -0,0 +1,14 @@
+#
+# file: migrations/0010.add_fkc_course_usage_table.py
+#
+from yoyo import step
+
+__depends__ = {'0009.add_meta_tables'}
+
+step('''
+        ALTER TABLE canvas_course_usage
+        ADD CONSTRAINT fk_canvas_course_usage_id
```
ssciolla, line 10: So the way I've been naming constraints is to reference what primary key you're linking to, so I think I would have done `fk_course_id`? What do you think?

pushyamig: That's what I did, but it showed an error saying the key already exists probably you used it for enrollment table. so I changed to this

ssciolla: Hm, yeah, I wondered about that. I think that's fine this way for now, pretty clear what you mean, and if there's confusion, there's the constraint itself which tells you what's going on.

PR #45, opened 2020-04-01
[DRAFT] integrate MiVideo data extraction programs

lsloan:
MiVideo programs temporary location.

I'm committing my MiVideo data extraction programs here for now to make sure they're checked in.  I will work on integrating this with the rest of the project's code.

Issue #66, opened 2020-04-07
Create DEFAULT_APPEND_TABLES, so we don't have to define metadata tables in env file


ssciolla:
I feel like we should have DEFAULT_APPEND_TABLES that we don't have to define in the config. Are we ever going to want these dropped automatically by the script?

like 
```
DEFAULT_APPEND_TABLES = ['job_run', "data_source_status"]
APPEND_TABLE_NAMES = ENV.get('APPEND_TABLE_NAMES', []).extend(DEFAULT_APPEND_TABLES)
```

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/course-inventory/pull/62_

ssciolla:
This is no longer relevant since resolving issue #72, so I'm removing this from the 2020.01.01 project.

Issue #29, opened 2020-03-30
Switch configuration from JSON to something else


jonespm:
I used YAML previously to configure the online meetings, we should switch the config for this over to using that. The biggest advantage would be being able to add comments to the config. In addition pasting in API keys and strings that span multiple lines is easier with YAML since they don't need to be quoted.

This can convert over the sample really quickly, then we just need to populate the production and make a few minor code changes to read the YAML instead of the JSON. 

https://www.json2yaml.com/

In addition to this, move the comments from the README.md related to these variables into the YAML file. 

In this article ["Why JSON isn't a good configuration language"](https://www.lucidchart.com/techblog/2018/07/16/why-json-isnt-a-good-configuration-language/) they mention altenratives like TOML, HOCON, HJSON. I like the idea of using something like HJSON or YAML because it would still be compatible with the files people are currently using.

They also link to an article [YAML: probably not so great after all](https://www.arp242.net/yaml-config.html) which is worth a consideration.

pushyamig:
From the master branch, I was able to test all the process ["COURSE_INVENTORY", "MIVIDEO", "CANVAS_ZOOM_MEETINGS"], with .hjson file. I will make it pass

PR #76, opened 2020-04-08
integrate MiVideo usage data

### LINKED ISSUE #65: integrate MiVideo usage data
Integrate the data showing the usage of MiVideo media in Canvas.  That is, the data that shows how many `MediaEvent`→`Started` Caliper events are found in UDP per course ID per hour.

The MiVideo media creation data will be covered in #92.

lsloan:
> ###### **_Note:_** This PR replaces #45.  I had created the earlier PR from a branch without an issue number in the name.  I then created the issue, #65, and renamed my branch correspondingly.  However, I forgot that the head branch cannot be changed in open PRs.  I will close the old PR soon after this new one opens.

Resolves #65.

Integrate the standalone MiVideo data extraction program into the course-inventory application.

## Requested changes
(Since 2020-04-16)

* [x] change migration docstring – https://github.com/tl-its-umich-edu/course-inventory/pull/76#discussion_r409686818
* [x] change DB column `course_id` to `INTEGER` – https://github.com/tl-its-umich-edu/course-inventory/pull/76#discussion_r409713445
* [x] change `LOG_LEVEL` type – https://github.com/tl-its-umich-edu/course-inventory/pull/76#discussion_r409687441
* [x] make UDP data source more specific – https://github.com/tl-its-umich-edu/course-inventory/pull/76#discussion_r409693281
* [x] tweak `env_blank.json` – https://github.com/tl-its-umich-edu/course-inventory/pull/76#discussion_r409698945
* [x] check if `lastTime` could be string `'NONE'` – https://github.com/tl-its-umich-edu/course-inventory/pull/76#discussion_r409712498
* ~[ ] determine whether `DBCreator..set_up()` necessary~ – https://github.com/tl-its-umich-edu/course-inventory/pull/76#discussion_r409707623


environ.py
```diff
@@ -1,25 +1,28 @@
 # standard libraries
```
lsloan, line 1: The more I use it, the more I think the `environ` module should be renamed `configuration`.  Because it doesn't represent the actual environment.  It's a combination of the environment and configuration settings.

ssciolla: I'm opening to discussing this, though I am not a fan of `configuration`, particularly. Please open a separate issue.

lsloan: OK.  See #94.

mivideo/.flake8
```diff
@@ -0,0 +1,7 @@
+[flake8]
```
ssciolla, line 1: I really don't want to hold up merging this for any style-based disagreements, but I suppose we should start the discussion. Are you planning on maintaining separate style guidelines for your job?

environ.py
```diff
@@ -1,25 +1,28 @@
...
-
-# Set up ENV
-
-config_path = os.path.join(ROOT_DIR, os.getenv('ENV_PATH', os.path.join('config', 'secrets', 'env.json')))
-
 try:
-    with open(config_path) as env_file:
-        ENV = json.loads(env_file.read())
+    with open(CONFIG_PATH) as env_file:
+        ENV: dict = json.loads(env_file.read())
```
ssciolla, line 18: Should probably be using `typing.Dict`.

environ.py
```diff
@@ -1,25 +1,28 @@
...
+import os
 from json.decoder import JSONDecodeError
+from typing import Union
 
+# Set up ENV
 # entry-level job modules need to be one-level beneath root
-ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
+ROOT_DIR: str = os.path.dirname(os.path.abspath(__file__))
+CONFIG_DIR: str = os.path.join(ROOT_DIR, os.getenv('ENV_DIR', os.path.join('config', 'secrets')))
+CONFIG_PATH: str = os.path.join(CONFIG_DIR, os.getenv('ENV_FILE', 'env.json'))
```
ssciolla, line 12: This looks like a fine solution to the problem we were discussing earlier. We'll need to update the README's OpenShift section, since `ENV_PATH` is no longer a thing.

environ.py
```diff
@@ -1,25 +1,28 @@
...
+    with open(CONFIG_PATH) as env_file:
+        ENV: dict = json.loads(env_file.read())
 except FileNotFoundError:
-    logger.error('Configuration file could not be found; please add env.json to the config directory.')
+    logger.error(
+        'Configuration file could not be found; please add env.json to the config directory.')
     ENV = None
 
-logging.basicConfig(level=ENV['LOG_LEVEL'])
+LOG_LEVEL: Union[str, int] = ENV.get('LOG_LEVEL', 0)
```
ssciolla, line 24: I'd prefer (to keep the typing simpler) if the default was `"INFO"` or `"DEBUG"`.

ssciolla:
Please update `env_blank.json` with the new key-value pair structures you are adding.

mivideo/mivideo_extract.py
```diff
@@ -0,0 +1,115 @@
...
+logger = logging.getLogger(__name__)
+
+SHAPE_ROWS: int = 0  # Index of row count in DataFrame.shape() array
+
+
+class MiVideoExtract(object):
+    udpKeyFilePath: str
+    credentials: service_account.Credentials
+    udpDb: bigquery.Client
+    appDb: DBCreator
```
ssciolla, line 26: I'm curious -- more for my own edification than any particular critique -- why are you using class variables over instance variables here?

mivideo/mivideo_extract.py
```diff
@@ -0,0 +1,115 @@
...
+
+        self.udpDb = bigquery.Client(
+            credentials=self.credentials,
+            project=self.credentials.project_id,
+        )
+
+        logger.debug(f'Connected to BigQuery project: "{self.udpDb.project}"')
+
+        DB_PARAMS = ENV['INVENTORY_DB']
+        APPEND_TABLE_NAMES = ENV.get('APPEND_TABLE_NAMES', ['job_run'])
```
ssciolla, line 46: The default should include `data_source_status`. Note though that this will become unnecessary whenever issue #66 is resolved (hopefully soon).

mivideo/mivideo_extract.py
```diff
@@ -0,0 +1,115 @@
...
+
+        self.appDb = DBCreator(DB_PARAMS, APPEND_TABLE_NAMES)
+        self.appDb.set_up()
+
+    def _readTableLastTime(self, tableName: str, tableColumnName: str) -> Union[str, None]:
+        lastTime: str = None
+
+        try:
+            sql: str = f'select max(t.{tableColumnName}) from {tableName} t'
```
ssciolla, line 55: Since you're really just using hard-coded values in `run` and this is only used once, it might just be better to use them here in the string, to avoid any SQL injection warnings.

ssciolla: If you need to use this for other processes you plan to add, not sure we have a great solution to the SQL injection warnings. I've been trying to deal with this too, and using parameters for database and table names with MySQL seems to be a non-starter; see [this StackOverflow question](https://stackoverflow.com/questions/40249590/sqlalchemy-error-when-adding-parameter-to-string-sql-query)

mivideo/mivideo_extract.py
```diff
@@ -0,0 +1,115 @@
...
+
+            logger.debug(f'"{tableName}" - Saving to table...')
+
+            dfCourseEvents.to_sql(tableName, self.appDb.engine, if_exists='append', index=False)
+
+            logger.debug(f'"{tableName}" - Saved.')
+        else:
+            logger.debug(f'"{tableName}" - No rows returned.')
+
+        logger.debug(f'"{tableName}" - Procedure complete.')
```
ssciolla, line 100: Some of your log messages should probably be `info` level. Please review and use your discretion.

mivideo/queries.py
```diff
@@ -0,0 +1,46 @@
...
+            ) AS event_hour_utc,
+                 event_time
+        FROM
+        `udp-umich-prod`.event_store.events
+        WHERE
+            (
+            ed_app = 'https://aakaf.mivideo.it.umich.edu/caliper/info/app/KafEdApp'
+        OR ed_app = 'https://1038472-1.kaf.kaltura.com/caliper/info/app/KafEdApp'
+        )
+        AND event_time > TIMESTAMP('{lastTime}')
```
ssciolla, line 32: I think `google-cloud-bigquery` has parameter functionality, so we should probably use that for the SQL Injection reasons broached above. See this example from MyLA: https://github.com/tl-its-umich-edu/my-learning-analytics/blob/36a05c27eee83dd702d65666586ebebf0ab4cf0c/dashboard/cron.py#L270-L287

db/migrations/1138.add_mivideo_tables.py
```diff
@@ -0,0 +1,14 @@
+from yoyo import step
```
ssciolla, line 1: The migration number in the file name will need to be updated. Also, in recent migrations, we've been using templated comments at the top and specifying the previous migration as a dependency. Let me know if you have questions/concerns.

db/migrations/1138.add_mivideo_tables.py
```diff
@@ -0,0 +1,14 @@
...
+
+steps = [
+    step('''
+        CREATE TABLE IF NOT EXISTS mivideo_media_started_hourly (
+            id BIGINT AUTO_INCREMENT NOT NULL UNIQUE,
+            event_hour_utc VARCHAR(20) NOT NULL,
+            course_id BIGINT NOT NULL,
```
ssciolla, line 8: I'm assuming this is Unizin ID? We may want to re-produce the Canvas course ID for simpler joining.

mivideo/mivideo_extract.py
ssciolla: If it's completely necessary, I think we should just make sure we're validating inputs where appropriate.

db/migrations/1138.add_mivideo_tables.py
lsloan: Questions:

1. I wasn't sure how many migrations there would be between the time I started working on the MiVideo implementation and the time my PR would get merged.  So I chose a sufficiently high number (`1138`).  Should I switch it to `0012` now?  Is there a convention I can use to reserve that number for the migration in my PR?
2. What is the significance of comments like `# file: migrations/0011.add_course_sis_ids.py`?  Is it necessary?  Isn't the file's own name sufficient to express this information?
3. I didn't think my implementation's migration depended on any earlier migrations, but thinking about it now, I guess it depends on `0009.add_meta_tables.py` for `job_run`.  Does that make sense?
4. Should I try to add some relationships between `mivideo_media_started_hourly` and other tables, like `course`?  That would also make dependencies between the jobs.  I'd no longer be able to run the MiVideo job without running the job that populates `course`, too.

ssciolla: Will get these answered after my meeting.

ssciolla: I would not add any relationships at this point, since this is an append table and course is not.

db/migrations/1138.add_mivideo_tables.py
lsloan: When you say "Unizin ID", I guess you mean the ID from UDW, with the `17700000000000000` base ID.  This is not that.  It doesn't use the base ID.

`course_id` contains the Canvas course ID, as shown in the Caliper events' `$.object.extensions.kaf:course_id` properties.  These are the same course IDs that appear in URLs used to access the course in Canvas.  That is, in the URL `https://umich.instructure.com/courses/332500`, it is `332500`.

I think we definitely will want to join or make foreign key relationships with `course` and this table.  If the `course` table uses the longer IDs, then I guess we'll need to adjust.  I find the adding/subtracting of the base ID to be annoying.  I wish it were the same thing everywhere.

Would it be OK to add another column with the longer ID?

environ.py
lsloan: I'll switch it to `'INFO'`.

mivideo/mivideo_extract.py
lsloan: AFAIK, this is for typing purposes.  They aren't assigned values at this point, so they are not class variables.

That is, that was my intention.  Did I get the syntax wrong?

mivideo/mivideo_extract.py
lsloan: It'd be nice if `DBCreator` were a singleton that would be instantiated with `APPEND_TABLE_NAMES` only once in the running of this application.  That is, `run_jobs.py` would probably call it and all of the individual jobs' code could just use it.

mivideo/mivideo_extract.py
lsloan: Yeah, it's used only internally in one place for now, but there will be a second MiVideo table added soon.  Actually, the other table was to be part of this PR.  However, its data isn't being used by the dashboard now.  So, I moved it to a separate issue so I could get this PR merged faster without it.

Where are you seeing the SQL injection warnings?  I don't see them when my code runs.  Are you using a different code quality checker or is Codacy telling us about this?

environ.py
lsloan: ✅Fixed.

environ.py
lsloan: ✅Fixed.

mivideo/mivideo_extract.py
ssciolla: Oh, that makes more sense to me now. I guess I've just been doing my type hinting when defining the instance variable, like here for `Job`: https://github.com/tl-its-umich-edu/course-inventory/blob/29d3ff0df96b95cb4b785302330289545aaab176/run_jobs.py#L26-L34

db/migrations/1138.add_mivideo_tables.py
ssciolla: We're no using the `canvas_id` as primary keys for the Canvas data tables, so you shouldn't need to do anything related to this if that's what you're getting back. I agree the relationship should be defined formally at some point, but we need to more thinking about linking tables that accumulate data and others that are always completely dropped.

db/migrations/1138.add_mivideo_tables.py
lsloan: I've changed…

* …the name to `0012…`.
* …the dependency to use `0009…`.

mivideo/mivideo_extract.py
ssciolla: Yes, the SQL injection warnings are coming from Codacy. Their scan doesn't seem very fine-grained or consistent, but they seem to flag SQL-like strings that have concatenation or interpolation.

mivideo/mivideo_extract.py
lsloan: ✅Fixed.

mivideo/mivideo_extract.py
lsloan: ✅Fixed.

mivideo/mivideo_extract.py
lsloan: I can change it to that style, if that's preferred.

Since Python typing is embraced more often, I like the style I used because it reminds me of other languages that *require* declarations at that level.  It makes a nice reference to the properties that will be used in the class.

db/migrations/1138.add_mivideo_tables.py
ssciolla: 1) There is no convention for reserving migration numbers yet, nor do I think it's really that necessary to have one. It's fairly straightforward to resolve any conflicts, especially since you will have named your migration differently (meaning the merge won't obliterate your file) and then you will just need to bump the number.

2) There isn't really any significance to the comment, other than extra clarity. I think the format came from the `yoyo` documentation. I didn't start off using it, I think @pushyamig  did first.

3) I think we've been fairly lackadaisical about the use of `__depends__`.  Recently to keep things simple we've just been specifying the most recent migration as the only dependency, regardless of whether it touches any of the same tables. From a technical database-oriented perspective, I don't think your migration has any dependencies. If you want to be consistent, I'd just use the `0011` migration. I don't think the metadata table migration is a dependency of this one.

lsloan:
Re: https://github.com/tl-its-umich-edu/course-inventory/pull/76#issuecomment-613587288
> Please update `env_blank.json` with the new key-value pair structures you are adding.

✅ Fixed.

environ.py
lsloan: ✅Fixed.

mivideo/.flake8
lsloan: Well, you mentioned the potential conflict between my preference for camelCase and others' for snake_case, so I thought, "what the heck?"

environ.py
lsloan: ;gree-check; Fixed.

mivideo/mivideo_extract.py
ssciolla: Yeah, I think the `DBCreator` question could be raised in a separate issue. I could see each job entry function accepting a `DBCreator` object...

mivideo/mivideo_extract.py
ssciolla: I think it's okay to keep it the way you have it, if you prefer it that way. Only drawback is that newbies like me might confuse them for class variables.

mivideo/mivideo_extract.py
lsloan: I've addressed most of the Codacy issues, I think.  I wish I knew of a faster way to get Codacy to check the code again locally, without needing to check changes into the repo.

WRT its issues about SQL injection, from what I looked up about SQLAlchemy and prevention of injection, it looks like the suggestion is to use its model features.  I'm not sure whether we want to get into that at this time.  I'm not opposed to it, but it seems like it will take an extra chunk of time to do it.

ssciolla: Agreed, I also looked at that, but we'd essentially have to adopt the whole ORM I think for the project, which wouldn't be bad but would mean we'd have to remove `yoyo` and learn a new technology. At which point, why aren't are we just using Django, which we know?

config/env_blank.json
```diff
@@ -38,5 +38,12 @@
     "user": "",
     "password": ""
   },
-  "APPEND_TABLE_NAMES": ["job_run", "data_source_status"]
+  "APPEND_TABLE_NAMES": [
+    "job_run", "data_source_status",
+    "mivideo_media_started_hourly", "mivideo_media_creation"
```
ssciolla, line 43: Should probably only include the tables you're adding for the time being.

config/env_blank.json
```diff
@@ -38,5 +38,12 @@
...
+  ],
+  "mivideo": {
+    "/*  'umich-its-tl-reports-prod.json' can be found at https://umich.app.box.com/file/641207013676": "*/",
+    "service_account_json_filename": "umich-its-tl-reports-prod.json"
```
ssciolla, line 47: Even though Box is behind SSO, I feel like this is too much information to provide here. I would probably just explain in the configuration README table what values are expected here, linked to the Google documentation where appropriate, and make sure other team members are aware of where to find this file you reference.

db/migrations/0012.add_mivideo_tables.py
```diff
@@ -0,0 +1,16 @@
+from yoyo import step
+
+__depends__ = {'0009.add_meta_tables.py'}
```
ssciolla, line 3: Please see my other migration-related comments. I don't really think this dependency is accurate from a database perspective (though it certainly is from a code one). I would be comfortable either with you listing the previously committed migration, or not listing anything.

lsloan: OK.  I'll remove the dependency, then.  Since we've agreed that the MiVideo data doesn't need to have a formal relationship with the course data, and the dependency on `job_run` doesn't count, because it's programmatic only, removing it is the best choice.

mivideo/mivideo_extract.py
lsloan: After more reading, I've decided to move those type declarations out of the class and into `__init__()`.

The way I defined them does make them available as class variables, but since the rest of the code was always accessing them as instance variables, if the class variables had been assigned any values, they would have been ignored.  So, no big deal, really.  It was working the way I intended to.

However, this means that two allocations were made for each variable, in each of the class and instance scopes.  So it's a waste of memory, processing, etc.

config/env_blank.json
lsloan: 🤷

lsloan: ✅Fixed.

config/env_blank.json
lsloan: ✅Fixed.

db/migrations/1138.add_mivideo_tables.py
lsloan: ✅Fixed.

Not adding the comment with embedded file name.  It's superfluous.  Also, if the name of the file or the directory changes later, will someone seriously go and update all the comments?

mivideo/mivideo_extract.py
lsloan: Opened #99.

mivideo/mivideo_extract.py
lsloan: ✅Fixed.

db/migrations/1138.add_mivideo_tables.py
ssciolla: That's fine, I'm not insisting on. FWIW, the path in the comment shouldn't change, since `yoyo` basically requires that these files be stored in a directory called `migrations`.

mivideo/queries.py
lsloan: ✅Fixed.

lsloan:
@ssciolla & @jonespm I believe I've addressed all the concerns raised in your reviews and those raised by Codacy.

WRT Codacy, I'm ignoring some issues it has with…

* `README.md` – because it's about the formatting of the long table in the middle of the document.  I agree with its issues, but if I change just the rows it complains about, it would look silly without changing the rest of it.
* `queries.py` – it claims there are SQL injection vulnerabilities, but I've addressed that.

I'll ping you both for reviews, hopefully final and approval.

db/migrations/0012.add_mivideo_tables.py
```diff
@@ -0,0 +1,18 @@
+'''
+Migrations for MiVideo tables
+'''
```
ssciolla, line 3: Are you planning on editing this file in the future to add more tables? Asking because you're using plural tables here and in the file name. I think database changes need to be made incrementally in separate migrations for everything to work correctly.

environ.py
```diff
@@ -1,25 +1,28 @@
...
+        ENV: Dict = json.loads(env_file.read())
 except FileNotFoundError:
-    logger.error('Configuration file could not be found; please add env.json to the config directory.')
-    ENV = None
+    logger.error(
+        f'Configuration file could not be found; please add file "{CONFIG_PATH}".')
+    ENV = dict()
 
-logging.basicConfig(level=ENV['LOG_LEVEL'])
+LOG_LEVEL: Union[str, int] = ENV.get('LOG_LEVEL', 'INFO')
```
ssciolla, line 24: I think the type here can just be `str` now.

mivideo/mivideo_extract.py
```diff
@@ -0,0 +1,134 @@
...
+    def run(self) -> Sequence[Dict[str, Union[ValidDataSourceName, pd.Timestamp]]]:
+        """
+        :return: Sequence of result dictionaries, with ValidDataSourceName and last run timestamp
+        """
+
+        tableName = 'mivideo_media_started_hourly'
+
+        logger.info(f'"{tableName}" - Starting procedure...')
+
+        lastTime: Union[str, None] = self._readTableLastTime(tableName, 'event_time_utc_latest')
```
ssciolla, line 81: I'm okay with this for now, but if you abstract this more to the point where it's accepting arbitrary input (from `ENV` perhaps), we should probably verify the `tableName` is valid (perhaps using `self.appDB.get_table_names()`.

run_jobs.py
```diff
@@ -43,7 +43,9 @@ def create_metadata(self) -> None:
             'finished_at': [finished_at_dt]
         })
         job_run_df.to_sql('job_run', db_creator_obj.engine, if_exists='append', index=False)
-        logger.info(f'Inserted job_run record with finished_at value of {finished_at_dt}')
+        logger.info(
+            f'Inserted job_run record for job_name "{self.name}" '
+            f'with finished_at value of "{finished_at_dt}"')
```
ssciolla, line 48:
```suggestion
        logger.info(
            f'Inserted job_run record for job_name "{self.name}" ' +
            f'with finished_at value of "{finished_at_dt}"'
        )
```

vocab.py
```diff
@@ -23,4 +23,5 @@ class ValidDataSourceName(Enum):
     """
     
     CANVAS_API = auto()
+    UNIZIN_DATA_PLATFORM = auto()
```
ssciolla, line 26: Do you think this should reference BigQuery? If we need to use the context store in the future, things might get confusing.

config/env_blank.json
```diff
@@ -38,5 +38,11 @@
     "user": "",
     "password": ""
   },
-  "APPEND_TABLE_NAMES": ["job_run", "data_source_status"]
+  "APPEND_TABLE_NAMES": [
+    "job_run", "data_source_status",
+    "mivideo_media_started_hourly"
+  ],
+  "mivideo": {
```
ssciolla, line 45: Can we at least make the highest level variable `ALL_CAPS`?

ssciolla: Also, I was trying to keep the data fetching variables higher up and the output-related stuff at the bottom. If you're amenable, maybe this can go under the `ZOOM_CONFIG`?

mivideo/mivideo_extract.py
```diff
@@ -0,0 +1,134 @@
...
+
+        logger.info(f'Connected to BigQuery project: "{self.udpDb.project}"')
+
+        DB_PARAMS = ENV['INVENTORY_DB']
+        APPEND_TABLE_NAMES = ENV.get('APPEND_TABLE_NAMES', [
+            'job_run', 'data_source_status', 'mivideo_media_started_hourly',
+            'mivideo_media_creation'])
+
+        self.appDb: DBCreator = DBCreator(DB_PARAMS, APPEND_TABLE_NAMES)
+        self.appDb.set_up()
```
ssciolla, line 58: If you're just going to use `engine`, you shouldn't need to call `set_up`.

mivideo/mivideo_extract.py
```diff
@@ -0,0 +1,134 @@
...
+
+    def _readTableLastTime(self, tableName: str, tableColumnName: str) -> Union[str, None]:
+        lastTime: Union[str, None]
+
+        try:
+            sql: str = f'select max(t.{tableColumnName}) from {tableName} t'
+            result: ResultProxy = self.appDb.engine.execute(sql)
+            lastTime = str(result.fetchone()[0])
+        except(Exception):
+            lastTime = None
```
ssciolla, line 68: When I ran this for the first time, there was no exception caught and instead the value of `lastTime` was the string `"None"`, which led to an invalid timestamp being passed to the query. I fixed it by changing to the following:

```
        sql: str = f'select max(t.{tableColumnName}) from {tableName} t'
        result: ResultProxy = self.appDb.engine.execute(sql)
        lastTime = str(result.fetchone()[0])

        if lastTime == 'None':
            lastTime = None
```

db/migrations/0012.add_mivideo_tables.py
```diff
@@ -0,0 +1,18 @@
...
+
+from yoyo import step
+
+steps = [
+    step('''
+        CREATE TABLE IF NOT EXISTS mivideo_media_started_hourly (
+            id BIGINT AUTO_INCREMENT NOT NULL UNIQUE,
+            event_hour_utc VARCHAR(20) NOT NULL,
+            course_id BIGINT NOT NULL,
```
ssciolla, line 12: Could this be just `INTEGER`? I've been using `INTEGER` for most Canvas IDs elsewhere unless values get larger than 6 digits

run_jobs.py
lsloan: Not necessary.

vocab.py
lsloan: Not BigQuery.  Maybe `UNIZIN_DATA_PLATFORM_EVENTS`?

config/env_blank.json
lsloan: Thinking…

mivideo/mivideo_extract.py
lsloan: Oh?  I'll see…

mivideo/mivideo_extract.py
lsloan: Will check…

db/migrations/0012.add_mivideo_tables.py
lsloan: Can't they get quite big, like those beginning with `1770000…`?

db/migrations/0012.add_mivideo_tables.py
lsloan: Yes.

db/migrations/0012.add_mivideo_tables.py
ssciolla: Those are the Unizin IDs. I am only seeing the standard 6 or fewer Canvas IDs in your table.

vocab.py
ssciolla: I'll leave it up to you.

db/migrations/0012.add_mivideo_tables.py
ssciolla: I don't think you can edit this in the future and have it do what you want. If the migration is marked as added in a database, I don't think (not 100 percent sure) any changes you make to the file will be applied. You can test this if you want, but I'm pretty sure you'll need to create a separate migrations.

run_jobs.py
ssciolla: :(

lsloan:
@ssciolla, I took a very quick look at your comments yesterday, so I'll look at them more closely now.

If it's not convenient for me to address the nitpicks, I'll skip them.  There are enough important questions that need to be answered.

I didn't expect it to take this long to review and merge this PR.  We have various levels of management breathing down our necks right now to get everything done quickly.  I'd like to get this done so it's ready to deploy to production, because I want to stop manually running the MiVideo data exports to CSV every day.  Also, this PR, addresses half of the MiVideo extracts.  The next step that I need to get working on is the media creation history data.  That's not being used by the dashboard, but there is interest to start using it **_soon_**.

I say that not to imply we should compromise on code quality, but we need to improve somehow.  I think one of my problems with this review so far is that the comments are scattered all through the conversation.  It takes a significant amount of time for me to scroll back and forth to make sure I've not skipped anything.  I think all comments should be nested under a single changes-requested comment per reviewer, so they stay grouped together.  Alternatively, I may make a summary comment after receiving change requests that lists all the changes with checkboxes so I can keep track of what needs to be done.

ssciolla:
I apologize for the lengthiness of the review, it was not my intention to drag this on. I am just trying to do my job, to ensure this will integrate with the rest of the codebase and that it will work.

Some of my comments may be nitpick-y; if you find them to be so, please feel free to say you're not going to fix them and move on. For anything that is not an actual blocker but you want to fix, you are welcome to make it an issue and resolve it later. In terms of process, you are welcome to request that I follow a particular comment pattern when reviewing your code.

mivideo/.flake8
ssciolla: I'm okay with this (at least for now), marking as resolved.

lsloan:
Thanks for understanding, @ssciolla.  I appreciate your suggestion to move some requested changes to issues when it makes sense.

My request is that when anyone reviews code, if the review will contain more than one comment to request changes, please group them together under a single "request-changes" or "comment" parent review comment.

I don't mean to accuse you or anyone of making the reviews too lengthy.  I can probably improve the way I handle the requested changes, and that would make the process more efficient, too.

db/migrations/0012.add_mivideo_tables.py
lsloan: OK.  I thought the longer IDs were still desired, as mentioned in the earlier comment: https://github.com/tl-its-umich-edu/course-inventory/pull/76#discussion_r408346194

So, yeah, I can make this `INTEGER`.

ssciolla: No worries, I thought I explained in a follow-up to that comment that we aren't using the long 17-digit IDs anymore, so the six-digit Canvas IDs you have should be fine.

db/migrations/0012.add_mivideo_tables.py
lsloan: In that case, then yes, I'll tweak the docstring.

mivideo/mivideo_extract.py
lsloan: Yeah, I was thinking about my change later that added `str()` there.  It's probably more useful for the function to return a time-like object (or `None`) as it was rather than `str`.  I needed to use the return value as a string somewhere else, so I should have cast it as such at that point, not make this function return `str`.

lsloan: Just because I was doing a substitution into a SQL string doesn't mean that the value needed to be a string, too.  I changed everything to use `datetime`.

config/env_blank.json
lsloan: OK.  I've changed the case and the position.  

Some tools that clean up JSON may change the position of the various properties.  If we really want to keep the data source properties together and the output properties together, it'd be better to make `dataSource` and `output` objects and nest the current properties within them.

ssciolla: That makes sense. Let's defer resolving that until we figure out what's happening with issue #29.

mivideo/mivideo_extract.py
lsloan: It's needed.  When I commented out the `set_up()` call, I saw messages like this in the log:

```txt
course_inventory | DEBUG:google.api_core.retry:Retrying due to , sleeping 16.7s ...
```

It was repeated several times, with various random values in place of "16.7s".

When I uncommented the line, the code worked again.

The reason I thought I needed to call `set_up()` is because it calls `engine.connect()`.  So, we could either call that in the MiVideo code, or we could call `set_up()`.  (0.5 = ½)

ssciolla: Okay, yeah, sorry for the confusion. I saw you using `result: ResultProxy = self.appDb.engine.execute(sql)`, and I kind of thought you were bypassing the connection setup part. I created `set_up` and `tear_down` to initialize and close the `self.conn` object, but you aren't using that, so I wasn't sure if it was necessary. If it were me, I'd be calling `self.appDB.conn.execute(sql)` and using `tear_down` afterward.

I am not sure why you are seeing `google` errors, this seems completely unrelated.

ssciolla: Alternatively, you could use `pandas.read_sql` and `self.engine` for getting that last value, and not worry about `set_up` and `tear_down`.

lsloan:
I've resolved all change requests from your last review, @ssciolla.  At the time I made those changes, I saw Codacy still complains about possible SQL injection vulnerabilities.  I think that's very unlikely.  The code no longer makes string substitutions and the value of the parameter used is a `datetime` object, either from a previous DB query or instantiated in the code.  I think Codacy may raise this issue just to be safe.

At the moment, Codacy is in the process of reviewing the code again.  It's been several minutes and it's still not ready.  If it doesn't find anything disastrous in the next few minutes, I think it's OK.

I will ping you and @jonespm for a new review.

mivideo/mivideo_extract.py
ssciolla: @lsloan, FWIW, I tried and removing the `set_up` and had no problem getting this to run.

lsloan:
> In an ideal world, the migration file name would be re-named `add_mivideo_table`, instead of the current plural.

Thanks for pointing that out.  I forgot to change that.

Merging momentarily…

mivideo/mivideo_extract.py
lsloan: Good suggestions, but I think I'll leave it as it is for now.

Issue #78, opened 2020-04-08
Support authentication into Zoom


jonespm:
We either should use JWT token (which has a Python Library) or OAuth2 Token (which also has a python Library)

Currently we've been using a pre-generated JWT but we may need to switch to use the other token.

https://github.com/jpadilla/pyjwt
https://github.com/requests/requests-oauthlib

jonespm:
Closing this as I'm doing it as part of #77

Issue #31, opened 2020-03-30
Get Canvas course & meetings code to use the database


jonespm:
Rather than writing files to the filesystem, we should be using the database in this project.
This will require
- [ ] Generate required migrations
- [ ] Adding to the migrations
- [ ] Deleting and adding the values instead of to_csv (with pandas to_sql for now)
- [ ] The index for ZOOM_CONFIG should probably be made into an int and we should have a mapping table indicating what this came from.

pushyamig:
I am assigning this issue to myself. I will take this up once I have a PR open for #25

zqian:
I removed the BlueJeans label from this ticket, since BJ is responsible for the BJ report now.

@jonespm please validate whether this issue is still valid given the Zoom report status.

jonespm:
Yeah, the scope of this has changed to just be for the Canvas course & meeting tables. We can update the title.

PR #79, opened 2020-04-08
#77 - Add support for multiple Zoom instances

### LINKED ISSUE #77: Add support for multiple Zoom instances
We're going to need to extend the Zoom code to pull from multiple Zoom instances.

- [x] Add multiple instances to the config
- [x] Iterate through both instances
- [x] Add a extra column to the database for the new instance
- [x] Support JWT authentication (formerly #78)

jonespm:
Add multiple instances to the config
 Iterate through both instances
 Add a extra column to the database for the new instance
 Support JWT authentication (formerly #78)

Fixes #77.


lsloan:
Fixes #77.

jonespm:
Thanks, guess I didn't notice it didn't do that! Saw this issue. 
https://github.com/isaacs/github/issues/568

Also cool browser extension mentioned on there that does take care of that
https://github.com/sindresorhus/refined-github

README.md
```diff
@@ -79,15 +79,18 @@ This includes the creation of a configuration file called `env.json`. Complete t
...
     `UDW` | An object containing the necessary credential information for connecting to the Unizin Data Warehouse, where data will be pulled from.
     `CREATE_CSVS` | A boolean (`true` or `false`) indicating whether CSVs should be generated by the execution.
     `INVENTORY_DB` | An object containing the necessary credential information for connecting to a MySQL database, where output data will be inserted.
     `APPEND_TABLE_NAMES` | An array of strings identifying tables that accumulate data and from which records should never be dropped programmatically.
+    `ZOOM_CONFIG` | An object containing the configuration for Zoom. You need a key for each object, this key is inserted into the DB (see below)
+    **ZOOM_CONFIG** |
+    `BASE_URL` | The base URL for calls to the Zoom API.
+    `API_KEY` | The API key to get the JWT token.
+    `API_SECRET` | The API secret to get the JWT token.
+    `EARLIEST_FROM` | The earliest date to retrieve zoom content from.
```
ssciolla, line 93: This looks okay for now. I suggested to @lsloan that we add another column for the nested level, but that doesn't work great with your key setup. Could you move this back up with the `SLEEP` param? Just trying to keep the data fetching stuff together. Also, could `EARLIEST_FROM` apply globally to all `ZOOM_CONFIG`s?

config/env_blank.json
```diff
@@ -13,9 +13,14 @@
   "CANVAS_URL": "",
   "CANVAS_TOKEN": "",
   "NUM_ASYNC_WORKERS": 8,
-  "ZOOM_TOKEN": "<ADD THIS TOKEN>",
-  "ZOOM_BASE_URL": "<ADD zoom.us URL>",
-  "ZOOM_EARLIEST_FROM": "YYYY-MM-DD",
+  "ZOOM_CONFIG": {
```
ssciolla, line 16: Yeah, I think this is the right place in the ordering for the `README`.

config/env_blank.json
```diff
@@ -13,9 +13,14 @@
...
+    "1": {
```
ssciolla, line 17: Could this be an `int`?

ssciolla: Also, what do you think the likelihood is of someone just mixing these up and putting one account that should be "1" as "2"?

online_meetings/report.py
```diff
@@ -83,31 +79,47 @@ def get_total_page_count(url: str, headers: Dict[str, Union[str, int]], params:
...
-            total_list.extend(zoom_loop(url, headers, json_attribute_name, dict(params), page_token))
-    else:
-        total_list.extend(zoom_loop(url, headers, json_attribute_name, dict(params), page_token))
+
+    for zoom_key, zoom_config in ENV["ZOOM_CONFIG"].items():
+        logger.info(f"Starting zoom pull for instance {zoom_key}")
+        url = zoom_config["BASE_URL"] + api_url
+        token = jwt.encode(
+            {'iss': zoom_config["API_KEY"],
+             'exp': datetime.utcnow() + timedelta(minutes=30)
```
ssciolla, line 100: Assuming I'm interpreting this correctly, would we ever need more than 30 minutes for an account run?

config/env_blank.json
jonespm: Yeah I wanted to do a table within a table (or something) in the README. We didn't have anything similar. If it was YAML we could have just commented right here! ;)

config/env_blank.json
jonespm: The number and order doesn't matter, this is the value that goes into a column in the database. I didn't know what Chris wanted so it's currently a string. If could be "um-zoom" or "um-zoom-health" as a VARCHAR unless we're doing a mapping.

online_meetings/report.py
jonespm: Yeah that's a good point. I think we either need a token refresh function or a longer time. Currently it's taking 20 minutes and if it went beyond that this would just stop running. I'll look at that as it might be more useful in the future too.

config/env_blank.json
ssciolla: Yeah, I'm concerned that someone will run it one way, and then someone else will do it and run it another.

README.md
jonespm: Yeah I'm not sure how best to format what's essentially needing to be a nested table. 

EARLIEST_FROM could apply globally but I actually think that would be more work, and I can see wanting the ability to configure these independently.

ssciolla: Fair enough on the `EARLIEST_FROM`, just thought I'd ask. Seemed to me like we'd want them to be in sync, and keeping them separate might introduce possible inconsistencies.

jonespm: I feel like the DEFAULT_SLEEP_TIME could be used by other API calls too so it's not really specific to Zoom. I'm not sure how it would look since these values only apply to Zoom. Hmm. . .

config/env_blank.json
ssciolla: As I think I've said before, I'm fine with switching to YAML, but having a separate resource still seems valuable to me.

requirements.txt
```diff
@@ -10,3 +10,4 @@ canvasapi==0.15.0
 beautifulsoup4==4.8.2
 dateparser==0.7.4
 pyyaml==5.3.1
+pyjwt
```
ssciolla, line 13: Let's pin this?

ssciolla: Also, I wouldn't complain if you combined with the others and alphabetized :)

README.md
ssciolla: I'm thinking `ZOOM_EARLIEST_FROM`.

config/env_blank.json
jonespm: I think this will need to be a separate ticket, probably can be done on #31.
 
When we create the DB tables for this we should probably have a mapping table. I feel like these are fine as an int and probably will need some way to define the mapping (probably via migration), just waiting to hear back from Chris.

requirements.txt
jonespm: Yeah, I remembered I wanted to do this last night but it was late when this finished, thanks.

online_meetings/report.py
```diff
@@ -83,31 +79,47 @@ def get_total_page_count(url: str, headers: Dict[str, Union[str, int]], params:
...
+             }, zoom_config["API_SECRET"], algorithm='HS256')
+        headers = {
+            # Need to decode https://github.com/jpadilla/pyjwt/issues/391
+            "Authorization": f"Bearer {token.decode('utf-8')}"
+        }
+        if use_date and "EARLIEST_FROM" in zoom_config:
+            early_date = dateparser.parse(zoom_config["EARLIEST_FROM"]).date()
+            # Only use the date as a paramter
```
ssciolla, line 108:
```suggestion
            # Only use the date as a parameter
```

online_meetings/report.py
ssciolla: So, do each of them take 20 minutes?

online_meetings/report.py
```diff
@@ -83,31 +79,47 @@ def get_total_page_count(url: str, headers: Dict[str, Union[str, int]], params:
...
+            # Loop until yesterday (this will go until now() -1)
+            for i in range((datetime.now().date() - early_date).days):
+                param_date = early_date + timedelta(days=i)
+                params["from"] = str(param_date)
+                params["to"] = str(param_date)
+                logger.info(f"Pulling data from date {param_date}")
```
ssciolla, line 114:
```suggestion
                logger.info(f"Pulling data for date {param_date}")
```

online_meetings/report.py
jonespm: No, the whole process is. I don't have that timed down to the specific run. The second zoom is like 10% less than the first so I'd guess it's 10% of the total.

README.md
jonespm: Well if it's in the block EARLIEST_FROM under Zoom makes sense. If we had an EARLIEST_FROM apply to all configs then we could put it on the top level of ZOOM_CONFIG and still call it the same, but this seems fine and is going to change when we only have to query just the updated data too.

online_meetings/report.py
jonespm: I don't think the token refresh will be hard to implement. We can even pro-actively do it without getting expired response back from the server. Though really I think the expiration time could also just be set to be much longer and Zoom doesn't care. Like we've had a JWT token that's had over a week expiration.

config/env_blank.json
jonespm: I added this as a task to #31

online_meetings/report.py
jonespm: Okay I rewrote this entire method to use a new class ZoomJWT. This when passed to requests will take care of setting the header and also doing the refresh if needed. This is now done as close as possible to the actual calls and all even more easily swappable.

jonespm:
Thanks for the feedback, upon further research I notice requests can take a parameter for a custom auth. I created one (since JWT wasn't an option) and it wasn't much work as it was what I already had. This gets the request really close to the call. This seems like a really cool idea and should incorporate it other places. 

I also enabled mypy in .vscode/settings.json when I was going through this and found a number of errors in the code I was working on. There might be issues in other code that wasn't addressed here.

README.md
ssciolla: Still would appreciate if this got moved up to where it was.

online_meetings/report.py
```diff
@@ -1,17 +1,19 @@
 import requests
+from requests.auth import AuthBase
 # standard libraries
 import yaml
 import json
 import logging
 import os
 
-from typing import Dict, Union
+from typing import Dict, Union, List
```
ssciolla, line 9: I've been using `Sequence` for lists, which it seems like `typing` recommends doing. See [the docs](https://docs.python.org/3/library/typing.html#typing.List`)

online_meetings/report.py
```diff
@@ -29,18 +31,18 @@
...
-}
-
-ZOOM_BASE_URL = ENV.get('ZOOM_BASE_URL', "")
 DEFAULT_SLEEP_TIME = ENV.get('DEFAULT_SLEEP_TIME', 10)
 
 
-def get_request_retry(url: str, headers: Dict[str, Union[str, int]],
+def requests_zoom_auth(url: str, auth: AuthBase,
+                       params: Dict[str, Union[str, int]]) -> requests.Response:
+    return requests.request("GET", url, params=params, auth=auth)
```
ssciolla, line 39: This is okay, though I'm not sure there's a lot gained by extracting as a function.

online_meetings/report.py
```diff
@@ -151,7 +163,10 @@ def zoom_loop(url: str, headers: Dict[str, Union[str, int]], json_attribute_name
                 page_token = results.get("next_page_token")
                 params["next_page_token"] = page_token
             elif params.get("page_number"):
-                params["page_number"] += 1
+                if (isinstance(params["page_number"], int)):
+                    params["page_number"] += 1
+                else:
+                    raise TypeError("Could not increment page number as it not an int")
```
ssciolla, line 169: Is this just some extra error handling, or were you responding to some particular case?

online_meetings/requests_zoom_jwt.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @property
+    def token(self) -> str:
+        """Check if the token is expired, if so generate a new token and return the token either way.
+        
+        :return: Returning the token value as a string
+        :rtype: str
+        """
+        # If token is close to expiring just re-create
+        if datetime.utcnow() > self.expiration - timedelta(seconds=30):
```
ssciolla, line 21: Nitpick: Might want to wrap `self.expiration - ...` in parentheses, just for clarity.

online_meetings/requests_zoom_jwt.py
```diff
@@ -0,0 +1,78 @@
...
+    def expiration(self, value: datetime):
+        self._expiration = value
+
+    def __init__(self, api_key: str, api_secret: str, alg="HS256", exp_seconds: int = 60):
+        """ Generates a JWT Token for Zoom
+        :param api_key: api key to use
+        :type api_key: str
+        :param api_secret: secret to use
+        :type api_secret: str
+        :param exp_seconds: How long to expire the token.
```
ssciolla, line 44:
```suggestion
        :param exp_seconds: How long the token should last before expiring
```

online_meetings/report.py
```diff
@@ -83,31 +84,42 @@ def get_total_page_count(url: str, headers: Dict[str, Union[str, int]], params:
...
-            total_list.extend(zoom_loop(url, headers, json_attribute_name, dict(params), page_token))
-    else:
-        total_list.extend(zoom_loop(url, headers, json_attribute_name, dict(params), page_token))
+
+    for zoom_key, zoom_config in ENV["ZOOM_CONFIG"].items():
+        zoom_list = []
+        logger.info(f"Starting zoom pull for instance {zoom_key}")
+        url = zoom_config["BASE_URL"] + api_url
+
+        auth = ZoomJWT(zoom_config["API_KEY"], zoom_config["API_SECRET"], exp_seconds=60 * 10)
```
ssciolla, line 105: Could we just make the default 600?

online_meetings/requests_zoom_jwt.py
```diff
@@ -0,0 +1,78 @@
...
+        self.expiration = datetime.utcnow() + timedelta(seconds=self.exp_seconds)
+        token = jwt.encode(
+                {'iss': self.api_key,
+                 'exp': self.expiration
+                 }, self.api_secret, algorithm=self.alg)
+        # Need to decode https://github.com/jpadilla/pyjwt/issues/391
+        return token.decode('utf-8')
+
+    def __call__(self, request: requests.PreparedRequest) -> requests.PreparedRequest:
+        """ For requests library to call
```
ssciolla, line 71: Was there any specific documentation that you referred to for this?

online_meetings/report.py
```diff
@@ -1,17 +1,19 @@
 import requests
+from requests.auth import AuthBase
```
ssciolla, line 2: Nitpick: these imports could be tidied up a bit.

online_meetings/requests_zoom_jwt.py
ssciolla: Also, might not hurt to have a debug logging message here for visibility on the token re-creation.

online_meetings/report.py
ssciolla: I'm playing around a bit with `mypy`, it's possible it wants `List` here, so up to your discretion.

README.md
jonespm: Don't you think it would be confusing? I don't think with Markdown we can add documentation for json objects and lists all in the same table as additional rows and not have it be confusing where those values apply to. I've haven't found any great example of documenting json in Markdown and if it does exist it won't be as a single table. I think anything this would involve us moving to readthedocs.io and having a Sphinx style documentation. 

I don't believe JSON makes a good configuration and I also don't think we should have this documention taking up space in the README (even if we're the only people looking at it). I feel like I've spent too much time trying to figure out a way to make this look good and understandable.

online_meetings/report.py
jonespm: Yeah, probably not anymore, I used to have not have "auth" but instead a method that generates this class. I realized that was wrong. I'll revert this separate function as it is unnecessary now that auth is passed in.

online_meetings/report.py
jonespm: With the mypy since it's `Union[str, int]` It needs a specific isinstance check or it flags it as an error. There's little to no chance of having a non-int case for this key but if it does come up I figure I'd throw the exception rather than continue on.

online_meetings/requests_zoom_jwt.py
jonespm: Thanks I wanted to add a little more debug to here, I'll add that.

online_meetings/report.py
jonespm: Yup!

README.md
ssciolla: Okay, don't worry about it anymore then, it's fine like it is for now, we can revisit when the change to YAML happens.

online_meetings/report.py
jonespm: Okay, doing this

jonespm: We should *probably* use some auto formatter. It doesn't look like Black covers this yet. This is probably tricky because PEP8 recommends you order it by

* Standard Libraries
* 3rd party libraries
* Local Libraries

Looks like isort might take care of this! https://pypi.org/project/isort/ I ran it on all the code in online_meetings.

online_meetings/requests_zoom_jwt.py
jonespm: It extends AuthBase, so it's part of that base classes documentation. I added that info to the header

online_meetings/report.py
ssciolla: Yeah, I follow that ordering and label as such, though I didn't know PEP8 recommended that, I don't think. Implementing an auto-formatter could be a separate issue.

README.md
jonespm: Yeah, in looking up the best way to do this I also saw a some negatives about YAML so that may need more research too. :)

jonespm:
Ok I think I got this all fixed up now. I also changed the ZOOM_CONFIG to a list rather than an object/dict as Chris is okay with just using an indexed variable. We'll have to in a future task (when we do the database) have a mapping somehow for this list item into a type in the database.

ssciolla:
> Ok I think I got this all fixed up now. I also changed the ZOOM_CONFIG to a list rather than an object/dict as Chris is okay with just using an indexed variable. We'll have to in a future task (when we do the database) have a mapping somehow for this list item into a type in the database.

Sounds fine. There are a couple docstring related Codacy issues, but not a bit deal to fix right now. Please feel free to merge.

jonespm:
Yeah, I fixed those two, I'm just giving it one more run to make sure the csv looks good after this config change and I'll commit. Thanks for all the feedback. This one looked simple but it actually did a lot to get this code quality better for the next step.

PR #80, opened 2020-04-09
load BlueJeans data into database

### LINKED ISSUE #55: BlueJeans output as csv file, or better, store the data into db
This won't work in the container. We're probably going to have to create a storage volume, pass in a path via ENV and use that to write. (Like /tmp for instance)

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/course-inventory/pull/48_

zqian:
The PR is for insert the BlueJeans API report dataframe into DB:

- somehow the db migration file for adding BlueJeans table is deleted in recent checkins. Added it back with modifications of several columns
- updated the run_jobs.py for the right package name for this BlueJeans meetings job;
- update the bluejeans_report.py file to adapt to the db job definition

Resolves #55.


pushyamig:
@zqian can you resolve conflicts?

db/migrations/0011.add_bluejeans_table.py
```diff
@@ -0,0 +1,42 @@
...
+            meetingTitle VARCHAR(255) NULL,
+            meetingId varchar(50) NOT NULL,
+            userName VARCHAR(255) NOT NULL,
+            user_id VARCHAR(255) NOT NULL,
+            user_tags  VARCHAR(255),
+            startTime DATETIME,
+            endTime DATETIME,
+            email VARCHAR(255),
+            participantSeconds int,
+            joinDate DATETIME,
```
pushyamig, line 21: Can you be consistent with the table naming convention? Here is a quick guideline https://launchbylunch.com/posts/2014/Feb/16/sql-naming-conventions/#naming-conventions

zqian: > @zqian can you resolve conflicts?

Just did!

db/migrations/0011.add_bluejeans_table.py
```diff
@@ -0,0 +1,42 @@
+#
+# file: migrations/0011.add_bluejeans_table.py
+#
+from yoyo import step
+
+__depends__ = {'0010.add_fkc_course_usage_table'}
```
pushyamig, line 6: The latest migration file from the master is `011.add_course_sis_ids.py` and you migration should be based on that
https://github.com/tl-its-umich-edu/course-inventory/blob/master/db/migrations/0011.add_course_sis_ids.py

Also, change the numbering on this file to 0012 something like that

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
+        payload = {}
+        headers = {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json',
+            'Authorization': f'Bearer {token}'
+        }
+        r = requests.get(download_url, headers=headers)
+
+        # specifying the zip file name
+        zip_file_name = "./bj.zip"
```
pushyamig, line 156: will this file name is always going be like `bj.zip`? what is the purpose of adding the `./ ` to this? If I have to run this locally do I need to have this located physically on my machine?

pushyamig: If the file name is going to be same all the time then make it a constant

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
+def run_bluejeans_report() -> Sequence[Dict[str, Union[str, pd.Timestamp]]]:
+    logger.info("run_bluejeans_report")
+
+    # Record data source info for Canvas API
+    bluejeans_data_source = {
+        'data_source_name': 'BLUEJEANS_API',
+        'data_updated_at': pd.to_datetime(time.time(), unit='s', utc=True)
+    }
+
+    url = "https://api.bluejeans.com/oauth2/token?Client"
```
pushyamig, line 54: What is the difference between these 2 URL `indigo-api.bluejeans.com` and `api.bluejeans.com`. Both are being used what is the reason for the different domain name?

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
+
+    payload = "{ \"grant_type\": \"client_credentials\", \"client_id\": \"Metrics_BJN\", \"client_secret\": \"" + ENV.get('BLUEJEANS_CLIENT_SECRET') + "\"}"
```
pushyamig, line 56: The backward slashed around double quotes around the payload variable attributes are little glary to the eye. can you make it a dictionary and convert it to JSON?

pushyamig:
A lot of duplicate code with header info. It can be addressed by pulling that info to a separate utils python file and making the token configuration values and just return respective headers

db/migrations/0011.add_bluejeans_table.py
pushyamig: The table column name should be in format "lower_case_with_underscores"

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
-    download_file_path = ""
-    while not download_file_path:
-        # retrieve report job information including the file download path
-        jobs_url = f"https://indigo-api.bluejeans.com/v1/enterprise/{ENV.get('BLUEJEANS_ENTERPRISE_ID')}/indigo/jobs/list?userid={ENV.get('BLUEJEANS_USER_ID')}"
+
+    total_df = pd.DataFrame()
+
+    # TODO: the earliest time is hardcoded now
+    from_date = dateparser.parse(ENV.get('ZOOM_EARLIEST_FROM', '2020-03-31')).date()
+    to_date = dateparser.parse(ENV.get('ZOOM_EARLIEST_TO', '2020-04-09')).date()
```
pushyamig, line 86: This feels if these environment variables are used for Bluejean and dates align with ZOOM these names can be refactored as  `ONLINE_EARLIEST_FROM`  `ONLINE_EARLIEST_TO`. Also, please update these variables in `env_blank.json`

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
+            'Accept': 'application/json',
+            'Content-Type': 'application/json',
+            'Authorization': f'Bearer {token}'
+        }
+        r = requests.get(download_url, headers=headers)
+
+        # specifying the zip file name
+        zip_file_name = "./bj.zip"
+
+        with open(zip_file_name, 'wb') as f:
```
pushyamig, line 158: Can this code be in the try-catch block?

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
+            f.write(r.content)
+        
+        # opening the zip file in READ mode
+        with ZipFile(zip_file_name, 'r') as zip_file:
```
pushyamig, line 162: This code needs to be in Try-catch it raises `ValueError`

pushyamig:
print statements are used in the python file in a couple of places, replace with log statement

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
+
+    payload = "{ \"grant_type\": \"client_credentials\", \"client_id\": \"Metrics_BJN\", \"client_secret\": \"" + ENV.get('BLUEJEANS_CLIENT_SECRET') + "\"}"
+    logger.info(payload)
+    headers = {
+    'Accept': 'application/json',
+    'Content-Type': 'application/json'
+    }
+
+    token_expiration_time = datetime.now()
+    response = requests.request("POST", url, headers=headers, data=payload).text.encode('utf8')
```
pushyamig, line 64: This should be try-catch block

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
+    token_json = json.loads(response)
```
pushyamig, line 65: This also needs to be try-catch

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
+            # retrieve report job information including the file download path
+            jobs_url = f"https://indigo-api.bluejeans.com/v1/enterprise/{ENV.get('BLUEJEANS_ENTERPRISE_ID')}/indigo/jobs/list?userid={ENV.get('BLUEJEANS_USER_ID')}"
+            payload = {}
+            headers = {
+                'Accept': 'application/json',
+                'Content-Type': 'application/json',
+                'Authorization': f'Bearer {token}'
+            }
+            jobs_response = requests.request("GET", jobs_url, headers=headers, data=payload)
+            jobs_json = json.loads(jobs_response.text.encode('utf8'))
```
pushyamig, line 132: both L131-132 needs to be in try- catch

pushyamig:
I am going to Generalize, the code when making an API request and when loading response JSON should be in Try-catch block

pushyamig:
Also after making an HTTP call a check for a Success response code should be made. We shouldn't assume that what we get from the server is always in the HTTP 200 series. I see this pattern all over when making Http calls

db/migrations/0011.add_bluejeans_table.py
```diff
@@ -0,0 +1,42 @@
...
+            popId int,
+            userType VARCHAR(255),
+            billable bool,
+            moderatorLess bool,
+            start_time DATETIME,
+            total_highlights_created int,
+            smart_meeting bool,
+            transcription_used bool,
+            total_transcription_duration_minutes float,
+
```
ssciolla, line 36:
```suggestion
```

db/migrations/0011.add_bluejeans_table.py
```diff
@@ -0,0 +1,42 @@
+#
+# file: migrations/0011.add_bluejeans_table.py
+#
+from yoyo import step
+
+__depends__ = {'0010.add_fkc_course_usage_table'}
+steps = [
+    step('''
+        CREATE TABLE IF NOT EXISTS bluejeans_meetings
```
ssciolla, line 9: We've been using singular table names, so could we change this to `bluejeans_meeting`?

db/migrations/0011.add_bluejeans_table.py
ssciolla: Agree with @pushyamig here, half the names are in CamelCase and the other half aren't; the rest of the database uses the underscore appraoch. Also, it would be nice to be consistent with putting data types in all caps and having one space between the column name and the data type.

db/migrations/0011.add_bluejeans_table.py
```diff
@@ -0,0 +1,42 @@
...
+        (
+            meeting_uuid varchar(255) NOT NULL UNIQUE,
+            meetingTitle VARCHAR(255) NULL,
+            meetingId varchar(50) NOT NULL,
+            userName VARCHAR(255) NOT NULL,
+            user_id VARCHAR(255) NOT NULL,
+            user_tags  VARCHAR(255),
```
ssciolla, line 16: Are these user fields for the host?

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
-    # post the report request job
-    url = "https://indigo-api.bluejeans.com/v1/enterprise/" + str(ENV.get('BLUEJEANS_ENTERPRISE_ID')) + "/indigo/reportJob/meetings?filter=\
-    [{\"type\":\"date\",\"comparison\":\"lt\",\"value\":\"" + end.astimezone().isoformat() + "\",\"field\":\"end_time\"},\
-    {\"type\":\"date\",\"comparison\":\"lt\",\"value\":\"" + end.astimezone().isoformat() + "\",\"field\":\"start_time\"},\
-    {\"type\":\"date\",\"comparison\":\"gt\",\"value\":\"" + start.astimezone().isoformat() + "\",\"field\":\"start_time\"}]\
-    &fileName=meetings_30th-Mar-2020_30th-Mar-2020&userid=" + str(ENV.get("BLUEJEANS_USER_ID")) + "&app_name=command_center"
+# output settings
+CREATE_CSVS = ENV.get('CREATE_CSVS', False)
+INVENTORY_DB = ENV['INVENTORY_DB']
+logger.info(INVENTORY_DB)
```
ssciolla, line 42:
```suggestion
```

online_meetings/bluejeans_report.py
```diff
@@ -33,141 +36,175 @@
...
+APPEND_TABLE_NAMES = ENV.get('APPEND_TABLE_NAMES', ['job_run', 'data_source_status'])
+
+def run_bluejeans_report() -> Sequence[Dict[str, Union[str, pd.Timestamp]]]:
+    logger.info("run_bluejeans_report")
+
+    # Record data source info for Canvas API
```
ssciolla, line 48: The data source name needs to be fixed in the comment. Also, this should really be done after you've done all of the data fetching, not before.

config/env_blank.json
```diff
@@ -38,5 +38,17 @@
...
+  "BLUEJEANS_URL_TOKEN": "",
+  "BLUEJEANS_URL_USER_JOBS": "",
+  "BLUEJEANS_URL_REPORT": "",
+  "BLUEJEANS_URL_DOWNLOAD": "",
+  "BLUEJEANS_URL_DOWNLOAD_ZIP_NAME": "bj.zip",
+  "BLUEJEANS_URL_DOWNLOAD_ZIP_NAME_INNER_FILE": "./meetings_*.csv",
+  "BLUEJEANS_CLIENT_SECRET": "",
+  "BLUEJEANS_USER_ID": 1,
+  "BLUEJEANS_FROM_DATE": "2020-03-01",
+  "BLUEJEANS_TO_DATE": "2020-03-02",
```
pushyamig, line 51: This is fine but the usual pattern has been for Json object that encapsulates all the variables associated with that
```
"BLUEJEANS": {
  "URL_TOKEN":"323423",
  "URL_USER_JOBS": "",
 .....
}
```

pushyamig:
This is Optional as well, The practice around this codebase is to create a class called BlueJeansReport and add all the methods as part of the class methods instead of having an orphan function controlled from `run_bluejeans_report`

pushyamig:
I want to test this by running locally, it looks like a lot more variables got added to `env.json`. Now I see a bluejeans token that is not part before. I need bluejeans credentials somewhere inbox to test this PR locally

pushyamig:
I was able to run the process successfully, Bluejean seems to be appending to an existing table. I am not sure if that needs to be part of `APPEND_TABLE` logic.

pushyamig:
It looks like we are appending to the bluejeans table. and it always getting one-day data as for from this comment in the code `# Loop until yesterday (this will go until now() -1)`
Please make sure in case of failure to run the job the next run find all the missed one and does not introduce a duplicate error in the Database. 
Like the night run job fails and retry also fails, the next day when it runs it might get the next routine. We might be missing a day. The code needs to handle pulling all the data in case of exceptions

online_meetings/bluejeans_report.py
```diff
@@ -32,141 +37,267 @@
...
+        logger.info(start.astimezone().isoformat())
+        logger.info("end")
+        logger.info(end.astimezone().isoformat())
+        
+        # post the report request job
+        report_url = ENV["BLUEJEANS_URL_REPORT"] + "?filter=\
+                    [{\"type\":\"date\",\"comparison\":\"lt\",\"value\":\"" + end.astimezone().isoformat() + "\",\"field\":\"end_time\"},\
+                        {\"type\":\"date\",\"comparison\":\"lt\",\"value\":\"" + end.astimezone().isoformat() + "\",\"field\":\"start_time\"},\
+                        {\"type\":\"date\",\"comparison\":\"gt\",\"value\":\"" + start.astimezone().isoformat() + "\",\"field\":\"start_time\"}\
+                    ]&fileName=meetings_30th-Mar-2020_30th-Mar-2020&userid=" + str(ENV.get("BLUEJEANS_USER_ID")) + "&app_name=command_center"
```
lsloan, line 250: This code generates a string with a lot of whitespace included.  Like…

```txt
?filter=                    [{"type":"date","comparison":"lt","value":"2020-03-01T23:59:59-05:00","field":"end_time"},                        {"type":"date","comparison":"lt","value":"2020-03-01T23:59:59-05:00","field":"start_time"},                        {"type":"date","comparison":"gt","value":"2020-03-01T00:00:00-05:00","field":"start_time"}                    ]&fileName=meetings_30th-Mar-2020_30th-Mar-2020&userid=None&app_name=command_center
```

This URL construction could be a lot easier to read if you changed the outer quotes from double `"` to single `'` (it will prevent the need for backslashes `\` in the string) and break it into pieces.  You can also eliminate the need for backslashes at line breaks after the `+` sign by wrapping the whole string in parentheses…

```suggestion
        report_url = (
            ENV['BLUEJEANS_URL_REPORT'] +
            '?filter=[' +
            '{"type":"date","comparison":"lt","value":"' +
            end.astimezone().isoformat() +
            '","field":"end_time"},' +
            '{"type":"date","comparison":"lt","value":"' +
            end.astimezone().isoformat() +
            '","field":"start_time"},' +
            '{"type":"date","comparison":"gt","value":"' +
            start.astimezone().isoformat() +
            '","field":"start_time"}' +
            ']&fileName=meetings_30th-Mar-2020_30th-Mar-2020&userid=' +
            str(ENV.get('BLUEJEANS_USER_ID')) +
            '&app_name=command_center'
        )
```

I was going to suggest using Python format strings (`f'…'`), but the curly braces in your strings caused problems with that.

In general, I suggest you avoid using `"` anywhere in Python programs where `'` can be used instead.  This is an example of why.

An even better solution that this would be to use Python dictionaries to hold all the URL parameters and use a URL construction module to build the URL for you.

lsloan: @zqian, I think you overlooked this comment.

zqian:
Closing the PR now:

- BJ team now is responsible to report BJ related data
- this code has been shared with the BJ team

Issue #67, opened 2020-04-07
Use a smarter method with docker-compose to determine when the MySQL database is ready and remove hehe from log message


ssciolla:
Right now `run_jobs.py` just waits 30 seconds for the MySQL container to finish initializing, but really it only needs to do that on the first startup. Otherwise, it takes three seconds. MyLA has a better strategy for this. See Matt's comment below.

"I'm doing this in the MyLA start script with the nc command which should be installed by default. 
```
echo "Waiting for DB"
while ! nc -z ${MYSQL_HOST} ${MYSQL_PORT}; do   
  sleep 1 # wait 1 second before check again
done
```

I believe the we could have a shell script before this python or could just do this similar check in python with nc. (I didn't test this code below)

```
while not os.system(f"nc -z {os.getenv("MYSQL_HOST")} {os.getenv("MYSQL_PORT")}):
    time.sleep (1)
```"

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/course-inventory/pull/62_

lsloan:
This is pretty excellent.  I made a `FIXME` comment in my own copy of the code that this needs to be done.  I'm glad others are thinking the same way.

90% of the time I run the app with docker-compose, the DB service is running and ready for use, but the program waits the 30 seconds anyway.

ssciolla:
This was a problem I had identified before submitting the `run_jobs.py` PR, but @jonespm said we could resolve if it after we merged. Thus, PR #81.

PR #81, opened 2020-04-09
Use loop to make waiting for MySQL container more responsive (#67)

ssciolla:
This PR adds a while loop to the beginning of `run_jobs.py` that checks whether the MySQL container is ready every 3 seconds, which leads to less waiting when the container has already been initialized before (creating a `.data` directory). The PR aims to resolve issue #67.


run_jobs.py
```diff
@@ -118,15 +119,26 @@ def run_jobs(self) -> None:
...
+        waiting = True
+        while waiting:
+            time.sleep(3.0)
+            try:
+                db_creator_obj.set_up()
+                db_creator_obj.tear_down()
+                logger.info('MySQL caught up')
+                waiting = False
+            except sqlalchemy.exc.OperationalError:
+                logger.debug('Still waiting!')
```
ssciolla, line 137: See 5:17 in https://www.youtube.com/watch?v=gyDb4szpWmc.

zqian: We should make it a requirement for linking to YouTube in code reviews -:)

Maybe this PR will fix my problem of not able to connect to MySQL instance with docker-compose approach. I used virtualenv afterwards.

pushyamig: I always test my DB changes in docker-compose and I did not have the problem of connecting to DB, deleting and inserting records

ssciolla: Yeah, I'm happy to help you @zqian figure out any issues with your `docker-compose` setup. This was just to make it so you don't have to wait 30 seconds every time on start up.

lsloan: > See 5:17 in https://www.youtube.com/watch?v=gyDb4szpWmc.

sameAsItEverWas

Qu'est-ce que c'est?

PS: YouTube URLs can include the start time: https://youtu.be/gyDb4szpWmc?t=317

run_jobs.py
```diff
@@ -118,15 +119,26 @@ def run_jobs(self) -> None:
...
+    db_creator_obj = DBCreator(ENV['INVENTORY_DB'], ENV['APPEND_TABLE_NAMES'])
     how_started = os.environ.get('HOW_STARTED', None)
 
     if how_started == 'DOCKER_COMPOSE':
-        logger.info('Waiting for the MySQL turtle, hehe')
         # Wait for MySQL container to finish setting up
-        time.sleep(30.0)
+        logger.info('Waiting for the MySQL turtle')
+        waiting = True
+        while waiting:
```
lsloan, line 129: Rather than test for a Boolean `waiting` flag as the `while` loop condition,  check whether `db_creator_obj.conn is not None`.

Also, I think there should be a limit to the number of times it tries.

run_jobs.py
```diff
@@ -118,15 +119,26 @@ def run_jobs(self) -> None:
...
+            time.sleep(3.0)
+            try:
+                db_creator_obj.set_up()
+                db_creator_obj.tear_down()
```
lsloan, line 133: When changing the loop condition to test `db_creator_obj.conn`, also move this call to `db_creator_obj.tear_down()` from inside the loop to just outside the loop.  That's what will leave the `.conn` property as non-`None`.

ssciolla:
> This PR has approvals already, so I'll post my review as comments only.

This PR does not have approvals yet.

run_jobs.py
```diff
@@ -118,15 +119,26 @@ def run_jobs(self) -> None:
...
 
 if __name__ == '__main__':
+    db_creator_obj = DBCreator(ENV['INVENTORY_DB'], ENV['APPEND_TABLE_NAMES'])
     how_started = os.environ.get('HOW_STARTED', None)
 
     if how_started == 'DOCKER_COMPOSE':
-        logger.info('Waiting for the MySQL turtle, hehe')
         # Wait for MySQL container to finish setting up
-        time.sleep(30.0)
+        logger.info('Waiting for the MySQL turtle')
```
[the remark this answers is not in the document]
ssciolla: This isn't meant as a dig on turtles, just FYI. As a turtle enthusiast, maybe you'd prefer to see the word `turtle` a little more often?

ssciolla:
@lsloan, based on your good suggestion of limiting how long it checks the connection, I switched this to a for loop maxing out at two minutes, or a little longer, at which time the process exits (`sys.exit(1)`). Let me know what you think. I don't see a huge difference between closing the connection (tear_down) inside or outside the loop, but I could be missing something.

run_jobs.py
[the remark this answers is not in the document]
lsloan: My turtles, Henry, Hermione, and Harold, feel they're often the victims of stereotypes.  Henry, in particular, points out that after he's gotten a good amount of basking time, he can get up to quite a good pace, especially when he's chasing some prey.

He's asked me to present Exhibit A:

image

lsloan:
> > This PR has approvals already, so I'll post my review as comments only.
> 
> This PR does not have approvals yet.

Oh?  I thought I'd seen that it had one approval. 🤷

lsloan:
> @lsloan, based on your good suggestion of limiting how long it checks the connection, I switched this to a for loop maxing out at two minutes, or a little longer, at which time the process exits (`sys.exit(1)`). Let me know what you think. I don't see a huge difference between closing the connection (`tear_down`) inside or outside the loop, but I could be missing something.

My suggestion to move `tear_down` out of the loop was based on the assumption that it would still be a `while` loop that tests `conn` for `None`.  I presume (sorry, didn't look) that if `set_up` sets `conn`, then `tear_down` would return it back to `None`.  If it did that within the loop, then `conn` would always be `None`.

Since you've abandoned the `while` loop for `for`, that point is moot.

run_jobs.py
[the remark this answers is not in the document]
ssciolla: Done!

ssciolla:
Just FYI, I'm thinking about changing `tear_down` and `set_up` to `connect` and `close` for clarity in a separate PR also addressing the fluent interface and default append table suggestions I've had from you and @jonespm. See https://github.com/tl-its-umich-edu/course-inventory/compare/master...ssciolla:improve-db-creator

run_jobs.py
lsloan: Since the loop has changed from `while` to `for`, there's no need to move `db_creator_obj.tear_down()` out of the loop.

lsloan:
> Just FYI, I'm thinking about changing `tear_down` and `set_up` to `connect` and `close` for clarity in a separate PR also addressing the fluent interface and default append table suggestions I've had from you and @jonespm. See [master...ssciolla:improve-db-creator](https://github.com/tl-its-umich-edu/course-inventory/compare/master...ssciolla:improve-db-creator)

That would be excellent!  I've been thinking that `set_up` and `tear_down` were not great descriptions of those methods. I was thinking of suggesting `connect` and `close` instead, but I was hesitant because I thought you had more functionality in mind for those methods with their original names.

ssciolla:
@lsloan or anybody, any change we could get this fully reviewed and approved soon? I think this improves developers' quality of life, and I'm working on other changes that might conflict.

run_jobs.py
```diff
@@ -118,15 +119,30 @@ def run_jobs(self) -> None:
...
 
 if __name__ == '__main__':
+    db_creator_obj = DBCreator(ENV['INVENTORY_DB'], ENV['APPEND_TABLE_NAMES'])
     how_started = os.environ.get('HOW_STARTED', None)
 
     if how_started == 'DOCKER_COMPOSE':
-        logger.info('Waiting for the MySQL turtle, hehe')
         # Wait for MySQL container to finish setting up
-        time.sleep(30.0)
+        logger.info('Waiting for the MySQL snail')
```
lsloan, line 127: |Team|Score|
|---|---|
|🐌|0|
|🐢|1|


😆

pushyamig:
This code seems to have conflicts, other than that, this looks Good!

ssciolla:
I resolved the conflicts, and I'm merging. Thanks all

PR #73, opened 2020-04-08
Fixes #68 and fixes #44 - Fixup README based on markdown lint rules and code based on flake8

jonespm:
Line length of 120 for Python and markdown
Moved flake8 and vscode settings to the top level

PR #83, opened 2020-04-09
update Docker config to rapid development

### LINKED ISSUE #82: mount source code with docker-compose for rapid dev
Change `docker-compose.yml` to mount the project source code directly in `/app`, eliminating the need for Docker to do anything during its `COPY` step.  That will also eliminate the need for rebuilding the `job` service container every time after the source code is changed.  This will make the development code-test cycle much quicker.

lsloan:
As indicated in the issue, change `docker-compose.yml` to mount project source code in `/app` to avoid container rebuild delays while developing.  Update documentation.

Resolves #82.


README.md
```diff
@@ -139,10 +142,70 @@ mysql --user=ci_user --password=ci_pw
 Use `^C` to stop the running MySQL container,
 or -- if you used the detached flag `-d` with `docker-compose up` -- use `docker-compose down`.
 
-Data in the MySQL database will persist after the container is stopped. 
-The MySQL data is stored in a volume mapped to the `.data/` directory in the project. 
+Data in the MySQL database will persist after the container is stopped.
+The MySQL data is stored in a volume mapped to the `.data/` directory in the project.
```
lsloan, line 146: Please excuse this whitespace change.  I don't think it's significant to Markdown.

My git client didn't tell me about this difference, but GitHub notices it.  I tried to undo the change and commit it,so it wouldn't show up as a change in this PR, but my git client won't recognize my undoing.

README.md
```diff
@@ -139,10 +142,70 @@ mysql --user=ci_user --password=ci_pw
...
 
+
+
+##### A Typical Development Cycle With Docker
+
+1. Build images for all services…
+
+    ```
+    docker-compose build
+    ```
```
ssciolla, line 157: For consistency with @jonespm's PR (#73), we should probably use language flags (e.g. `bash`).

lsloan:
Yes.  This is one of the problems I had with MyLA.  It didn't do this, either.  I constantly had to rebuild the container.  That project needs a change similar to this.

Still, during the initial build, I don't know why Docker pauses for a while during the `COPY` step.  Is it just checking or is it copying everything back onto itself?  I will eventually try to run docker-compose in verbose or debug mode and find out what it's doing.  But I don't think that will hold back this PR, because it seems to be working well for development purposes and it's not intended for production.

I'll make that Markdown change you suggested.

README.md
lsloan: I'll use `sh` to be shell agnostic.

lsloan:
I've made the Markdown changes @ssciolla suggested.  I also fixed a couple other spacing issues.

# **_Is this PR ready to merge?_**

(Pfft!  I just noticed that my earlier commits haven't been signed.  Shame!  I'll fix that for future commits.)

ssciolla:
> # Is this PR ready to merge?

Curious to know @jonespm's opinion, but I think it looks good.

README.md
```diff
@@ -194,15 +259,15 @@ this README. However, a few details about how the job is configurd are provided
...
-      - name: ENV_FILE
-        value: /config/secrets/env_test.json
-      - name: USER
-        value: project_name
-  ```
+    - env:
+        - name: ENV_FILE
+            value: /config/secrets/env_test.json
+        - name: USER
+            value: project_name
```
ssciolla, line 269: Is this standard indentation for YAML?

lsloan: I'll check.  Codacy didn't complain.

lsloan: Maybe not.

One source (the search result highlighted by Google) recommends indentation of 2 spaces, not 4.  However, [they also say](https://docs.saltstack.com/en/master/topics/troubleshooting/yaml_idiosyncrasies.html), "but YAML will follow whatever indentation system that the individual file uses."

Furthermore, the [YAML 1.2 specifications](https://yaml.org/spec/) don't give a specific number of spaces, but say that it needs to be consistent within a single file.  Their examples show 2, 4, and sometimes even 3 space indentation.

My Markdown editor is set to 4 because that's what that markup expects.  That's the problem we get when embedding code in text.  A different editor might not have this problem, I suppose.

Since you seem to be concerned by 4-space indentation, I'll change the indentation of this code to 2 spaces (within the Markdown quotes) this time just to get this PR though review and merged.

ssciolla: I mostly didn't know why you were indenting those lines more than the `-` lines. I'm really not very experienced with YAML, so I'm not going to make a big deal about this.

lsloan:
I've change the indentation in the example YAML block to two spaces.

Also, if we're going to be very strict, we shouldn't refer to YAML files as `.yml`, they should be `.yaml`, according to the [yaml.org FAQ](https://yaml.org/faq.html).

Yes, I'm guilty of changing some of the `.yaml` references to `.yml`.  I did that to be consistent with the other preexisting occurrences of `.yml` in the file and I wasn't aware of yaml.org's recommendation.  Should we change all that now?

Also, is there an example YAML file that we should include in the project?

README.md
lsloan: I think we're both making nitpick-y review comments that need to be prefixed with "nit:" as suggested by the ["How to Make Good Code Reviews Better"](https://stackoverflow.blog/2019/09/30/how-to-make-good-code-reviews-better/) article @jonespm recommended recently.

ssciolla:
> Also, if we're going to be very strict, we shouldn't refer to YAML files as `.yml`, they should be `.yaml`, according to the [yaml.org FAQ](https://yaml.org/faq.html).
> 
> Yes, I'm guilty of changing some of the `.yaml` references to `.yml`. I did that to be consistent with the other preexisting occurrences of `.yml` in the file and I wasn't aware of yaml.org's recommendation. Should we change all that now?

This isn't a big deal I don't think, but if we want to set a standard, I would keep instances of `docker-compose.yml` the same, since that's what Docker uses. Any others can be `.yaml`.

> Also, is there an example YAML file that we should include in the project?

In my opinion, no. As I said in the README, I think the full OpenShift integration (which we won't be using for much longer anyway) is out of scope, and much of it we just wouldn't be able to fill for security reasons.

README.md
ssciolla: Yeah, I've started using "Nitpick:". But you can assume that, if I've already approved something and haven't gone back and requested changes, that you can merge without resolving any open questions.

README.md
jonespm: I commented on another issue (In MWrite I think) for the purposes of language sh/zsh/shell-script/bash [are all aliases for the same thing.](https://github.com/github/linguist/blob/40992ba7f86889f80dfed3ba95e11e1082200bad/lib/linguist/languages.yml#L4984) It doesn't matter which one you use in Markdown here.

README.md
```diff
@@ -180,7 +245,7 @@ this README. However, a few details about how the job is configurd are provided
 * The `env.json` file described in the **Configuration** section above needs to be made available to 
   running course-inventory containers via an OpenShift ConfigMap, a type of Resource. A volume containing the ConfigMap 
   should be mapped to the `config/secrets` subdirectory. These details will be specified in a configuration file
-  (.yaml) defining the pod.
```
jonespm, line 183: This currently does have .yaml as the extension. Maybe just remove it entirely or don't use an extension as it could be either .yaml or .yml.

FYI: [Is there an official extension for YAML files? Please use ".yaml" when possible.
](https://web.archive.org/web/20060924190202/https://yaml.org/faq.html) We aren't limited by 8.3 anymore. ;) This advice is unchanged since 2006.

lsloan: The current version has some references to `yaml` and some to `yml`.  Some of those are as filename extensions, and some are not.  I'll just update all the references to "YAML".

lsloan:
So, I've changed all "yml" references in `README.md` to "yaml".  I've also renamed `docker-compose.yml` to `docker-compose.yaml`.  Funny that renames aren't shown under the PR's "Files changed" tab.

@jonespm, I'm pinging you for a review because @ssciolla says he would feel more comfortable with your **_approval_** on this PR before we merge it.

README.md
lsloan: Cool.  Then `sh` is a good choice.  Thanks!

lsloan:
Thanks, @jonespm and @ssciolla for the reviews, feedback, and approvals!

Issue #75, opened 2020-04-08
Flake8 identified response_data as not being used in inventory.py


jonespm:
Was this method intending to return the response_data or just the response? This should be double checked and just do a json.loads if we're trying to get an exception. This seems like it could be a bug but assigning someone that knows more about this code.

https://github.com/tl-its-umich-edu/course-inventory/blob/079d22d4251e285236653bc065f46a03381c6a30/course_inventory/inventory.py#L57

ssciolla:
Yeah, I can change this. I figured it was pretty harmless to assign to something, but it's not necessary.

PR #85, opened 2020-04-09
Remove assignment to unused var, make other inventory changes for mypy (#75)

ssciolla:
This PR aims to satisfy a Codacy style warning about an unused variable by removing the assignment. In addition, some minor changes to code and type hints were made to fix `mypy` errors in `inventory.py`. The PR aims to resolve issue #75.

### DIFF
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -1,7 +1,7 @@
 # standard libraries
 import json, logging, os, time
 from json.decoder import JSONDecodeError
-from typing import Any, Dict, List, Optional, Sequence, Union
+from typing import Any, Dict, List, Sequence, Union
 
 # third-party libraries
 import pandas as pd
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -134,7 +134,7 @@ def gather_course_data_from_api(account_id: int, term_id: int) -> pd.DataFrame:
 
 # Function(s) - UDW
 
-def process_sis_id(orig_sis_id: str) -> Optional[int]:
+def process_sis_id(orig_sis_id: str) -> Union[int, None]:
     try:
         sis_id = int(orig_sis_id)
         return sis_id
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -1,7 +1,7 @@
 # standard libraries
 import json, logging, os, time
 from json.decoder import JSONDecodeError
-from typing import Dict, Sequence, Union
+from typing import Any, Dict, List, Optional, Sequence, Union
 
 # third-party libraries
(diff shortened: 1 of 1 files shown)


course_inventory/inventory.py
```diff
@@ -64,7 +64,7 @@ def make_request_using_api_utils(url: str, params: Dict[str, Union[str, int]] =
     return response
 
 
-def slim_down_course_data(course_data: Sequence[Dict]) -> Sequence[Dict]:
+def slim_down_course_data(course_data: List[Dict]) -> List[Dict]:
```
lsloan, line 67: Why the change from `Sequence` → `List`?

course_inventory/inventory.py
```diff
@@ -134,13 +134,13 @@ def gather_course_data_from_api(account_id: int, term_id: int) -> pd.DataFrame:
 
 # Function(s) - UDW
 
-def process_sis_id(orig_sis_id: str) -> Union[int, None]:
+def process_sis_id(orig_sis_id: str) -> Optional[int]:
```
lsloan, line 137: I considered using `Optional` in some of my code, but it made it look like a result value from the function was, well… optional.  I preferred `Union[…, None]` because it made it explicit that my function would return either the first type or `None`.

ssciolla: I can try switching back to that. I like `Union`, too.

course_inventory/inventory.py
ssciolla: I was getting an error from `mypy` about argument types not matching, and this fixed it. I will recreate the error and post here.

ssciolla: ```course_inventory/inventory.py:113: error: Unsupported left operand type for + ("Sequence[Dict[Any, Any]]")```

ssciolla: I guess where I'm using `+` between lists on 113, it doesn't like that the type hint is `Sequence`?

course_inventory/inventory.py
ssciolla: Done, I made the switch back here.

course_inventory/inventory.py
jonespm: From the readme. "Dict and List are aliases for the built-ins dict and list, respectively."

"Iterable, Sequence, and Mapping are generic types that correspond to Python protocols. For example, a str object or a List[str] object is valid when Iterable[str] or Sequence[str] is expected."

So it seems like the way Python duck typing works it will throw an error for these operands if it's not always possible to do the check. You can probably also check for isinstance(var, dict) and have mypy be good with it too later in the code. What you have here seems correct tho. Still learning this stuff.

PR #89, opened 2020-04-10
#88 enabling SSL to support connections within MiServer

### LINKED ISSUE #88: Support connecting from MIServer instance MiDatabase by enabling SSL
The new MiServer set for the course inventory project the Database connection string needs to be enabled with SSL so that it can connect with MiDatabase instance

pushyamig:
I was able to test this on all the following scenarios
1. Miserver instance: Tesla
2. Localhost mysql DB instance
3. From Docker compose build

Resolves #88.

### DIFF
--- a/db/db_creator.py
+++ b/db/db_creator.py
@@ -29,7 +29,7 @@ class DBCreator:
             f":{quote_plus(db_params['password'])}" +
             f"@{db_params['host']}" +
             f":{db_params['port']}" +
-            f"/{db_params['dbname']}?charset=utf8"
+            f"/{db_params['dbname']}?charset=utf8&ssl=true"
         )
         self.engine = create_engine(self.conn_str)
         self.append_table_names = append_table_names


ssciolla:
I don't think this is a bug, but an enhancement. We have not had to deal with SSL or certificates until now.

pushyamig:
If you want to test this PR on Tesla, these are the steps need to be followed. Docker is already installed system-wide so this shouldn't be a problem running the docker command
1. ssh uniqname@tesla.dsc.umich.edu, log in  with you umich password
2. You will end up in the home directory(pwd), git clone project, cd into the project
3. add the `config/secrets/env.json` to the desired folder using 
`scp env.json uniqname@tesla.dsc.umich.edu:/home/<uniqname>`
4. Build`docker build -t <build-name>`
5. Run: `docker run <build-name>`

lsloan:
@pushyamig, this PR has been approved, so maybe it's time to merge it and… :shipit:

pushyamig:
this fix was put in in support of MIserver Setup for the project, but whether to have that or not is being in Question? So this change shouldn't hurt anything if we decided to drop the idea of running the process from MiServer. I will merge for now

PR #93, opened 2020-04-13
Canvas Zoom Meetings no longer working 04/13/2020

### LINKED ISSUE #91: Canvas Zoom Meetings no longer working 04/13/2020
The Canvas Zoom Meetings script is no longer working.

Some research indicates Zoom added a value lti_scid as a query parameter and it doesn't work without this. We'll have to add this to get it working again.

jonespm:
Also some mypy cleanup for canvas_zoom_meetings

Fixes #91

### DIFF
--- a/online_meetings/canvas_zoom_meetings.py
+++ b/online_meetings/canvas_zoom_meetings.py
@@ -1,6 +1,5 @@
 # Script to get all sites where Zoom is visible and retrieve the meetings to generate a report
 
-import http.client
 import json
 import logging
 import math
@@ -8,7 +7,7 @@ import os
 import re
 import sys
 from datetime import datetime
-from typing import Dict
+from typing import Dict, Optional, List, Union
 
 import canvasapi
 import pandas as pd
@@ -27,12 +26,6 @@ except FileNotFoundError:
 LOG_LEVEL = ENV.get('LOG_LEVEL', 'DEBUG')
 logging.basicConfig(level=LOG_LEVEL)
 
-# These two lines enable debugging at httplib level (requests->urllib3->http.client)
-# You will see the REQUEST, including HEADERS and DATA, and RESPONSE with HEADERS but without DATA.
-# The only thing missing will be the response.body which is not logged.
-if LOG_LEVEL == logging.DEBUG:
-    http.client.HTTPConnection.debuglevel = 1
-
 # You must initialize logging, otherwise you'll not see debug output.
 logger = logging.getLogger(__name__)
 logger.setLevel(LOG_LEVEL)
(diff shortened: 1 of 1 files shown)


jonespm:
The main issue is it needed a new parameter of lti_scid. I did some refactoring to make it easier to pass these extra values in the future in.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -67,42 +72,61 @@ def get_zoom_json(self, page_num: int = 1) -> Dict:
             return zoom_json["result"]
         return None
 
+    def extract_from_js(self, key: str, text: str) -> Optional[str]:
```
ssciolla, line 75: Not sure if this should be a method if it doesn't require `self`.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -67,42 +72,61 @@ def get_zoom_json(self, page_num: int = 1) -> Dict:
...
             })
 
-            zoom_json = self.get_zoom_json(1)
+            zoom_json = self.get_zoom_json({'page': 1, 'lti_scid': scid})
             # The first call to zoom returns total and pageSize, get the total pages by dividing
             if zoom_json:
                 total = zoom_json["total"]
                 total_pages = math.ceil(total / zoom_json["pageSize"]) + 1
-            for page_num in range(1, total_pages):
+            for page in range(1, total_pages):
```
ssciolla, line 110: I like `page_num` a little better, but not a big deal, really.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -67,42 +72,61 @@ def get_zoom_json(self, page_num: int = 1) -> Dict:
...
-                        'meeting_number': meeting['meetingNumber'],
-                        'host_id': meeting['hostId'],
-                        'topic': meeting['topic'],
-                        'join_url': meeting['joinUrl'],
-                        'start_time': meeting['startTime'],
-                        'status': meeting['status'],
-                        'timezone': meeting['timezone']
-                    })
+                if page != 1:
+                    zoom_json = self.get_zoom_json({'page': page, 'lti_scid': scid})
```
ssciolla, line 113: Nitpick: I find this logic a little clunky, but it should work fine. I think in `inventory.py` I just wrote a parsing or crosswalk function so I could use it inside and outside the loop.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -67,42 +72,61 @@ def get_zoom_json(self, page_num: int = 1) -> Dict:
...
+        """Takes a javascript text and attempts to extract a key/value (May not work with everything)
+        
+        :param key: key to extract
+        :type key: str
+        :param text: text to search in
+        :type text: str
+        :return: The matching string or None if not found
+        :rtype: Optional[str]
+        """
+        pattern = re.search(f'{key}.*"(.*)"', text)
```
ssciolla, line 85: This seems to be doing what you want it to. I wish there was a way we could just parse the string as JSON, as that seems less error-prone than a somewhat-hard-to-read regular expression. However, looking at one example, it looks like there are some single quotes in the `<script>` block, so you might have to do some string cleaning before you could successfully parse it.

online_meetings/canvas_zoom_meetings.py
jonespm: Yeah, good point, I could use a @staticmethod annotation here.

online_meetings/canvas_zoom_meetings.py
jonespm: The only reason I changed it is because the parameter on the zoom API call was just *page* I think it's worth being consistent with that.

online_meetings/canvas_zoom_meetings.py
jonespm: Yeah, it is a little clunky. I could change it to **kwargs and maybe it'd be a little better than posting an dict of data?

online_meetings/canvas_zoom_meetings.py
jonespm: Yeah, I tried to research this a bit but the problem is that this isn't JSON this is Javascript coming back so it's not as parsable a language. I didn't find any great updated package that could parse javascript in Python and most SO suggestions just recommended using Regex. 

There's stuff out there like https://github.com/rspivak/slimit and https://github.com/PiotrDabkowski/pyjsparser to lex Javascript but they're out of date and probably too much. I would like to make this better but not spend too much time on it since I currently only need two specific keys. Maybe something to keep in mind.

jonespm:
Just making sure you're okay with the latest changes. I think it's a slight improvement but this could still be better.

online_meetings/canvas_zoom_meetings.py
ssciolla: No worries. In my own code, I prefer to distinguish somehow between scalars and objects in the variable name when possible.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -45,64 +38,96 @@
...
-                        'meeting_id': meeting['meetingId'],
-                        'meeting_number': meeting['meetingNumber'],
-                        'host_id': meeting['hostId'],
-                        'topic': meeting['topic'],
-                        'join_url': meeting['joinUrl'],
-                        'start_time': meeting['startTime'],
-                        'status': meeting['status'],
-                        'timezone': meeting['timezone']
-                    })
+                if page != 1:
```
ssciolla, line 113: Might have been some miscommunication here on my part. I was referring to his condition being clunky (in other words, that it would be clearer to process this first request outside the loop. Again, not a super big deal.

jonespm: Yeah, these Zoom API's are special since you need to do the first call to get the page size. It may be 0 and if it is we don't need to do anything but if it's > 0 we have to run this loop. I just figured to avoid the duplicate call to the API. Maybe this could be refactored but I think I'll leave it for the moment.

PR #97, opened 2020-04-14
Pull enumerations from run_jobs.py into separate vocab.py module (#96)

### LINKED ISSUE #96: Enums from run_jobs.py should be in a separate module
The `ValidJobName` and `ValidDataSourceName` enumerations should be pulled out of `run_jobs.py` (possibly to a `vocab.py` module) so they can be used by different jobs' entry modules, as well as `run_jobs.py`, without creating a circular dependency.

ssciolla:
This PR pulls out the enumerations from `run_jobs.py` into a separate `vocab.py` module so that both `run_jobs.py` and job-specific modules can reference the enumeration without introducing a circular dependency. It also changes the format of the `data_source` dictionaries returned from job entry functions so that the value associated with the key `data_source_name` is a member of `ValidDataSourceName`. This PR aims to resolve #96.

### DIFF
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -13,6 +13,7 @@ from umich_api.api_utils import ApiUtil
 # local libraries
 from db.db_creator import DBCreator
 from environ import ENV
+from vocab import ValidDataSourceName
 from .async_enroll_gatherer import AsyncEnrollGatherer
 from .canvas_course_usage import CanvasCourseUsage
 from .gql_queries import queries as QUERIES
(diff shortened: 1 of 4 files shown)


vocab.py
```diff
@@ -0,0 +1,22 @@
+# standard libraries
+from enum import auto, Enum
+
+
+# Enum(s)
+
+# Each job name should be defined in ValidJobName.
+# NAME_OF_JOB = 'path.to.method'
```
lsloan, line 8: What do you think of moving comments like these into docstrings?

course_inventory/inventory.py
```diff
@@ -180,7 +181,7 @@ def pull_sis_section_data_from_udw(section_ids: Sequence[int], conn: connection)
 
 # Entry point for run_jobs.py
 
-def run_course_inventory() -> Sequence[Dict[str, Union[str, pd.Timestamp]]]:
+def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.Timestamp]]]:
```
lsloan, line 184: I like that the return type is no longer a mysterious, undescribed string.

lsloan:
I changed the description's "resolve issue #96" to "resolve #96" to trigger GH's issue linking feature.

vocab.py
ssciolla: I moved them, let me know if you see any issues with them.

PR #101, opened 2020-04-16
Generate participant minutes for zoom meetings

### LINKED ISSUE #95: generate participantMinutes for Zoom meetings
Zoom meeting records does not have participantMinutes, vs BlueJeans data feed has that entry.

We can estimate the number by doing the following:

(end_time-start_time)*participants

jonespm:
Fixes #95

### DIFF
--- a/online_meetings/report.py
+++ b/online_meetings/report.py
@@ -36,7 +36,7 @@ def to_seconds(duration: str) -> int:
             total_seconds = total_seconds * 60 + int(second)
         # If there is an error just return 0
         except ValueError:
-            logger.warn(f"{str} is not a valid duration")
+            logger.warning(f"{str} is not a valid duration")
             return 0
     return total_seconds
 
--- a/online_meetings/report.py
+++ b/online_meetings/report.py
@@ -21,6 +21,26 @@ logging.basicConfig(level=ENV.get('LOG_LEVEL', 'DEBUG'))
 DEFAULT_SLEEP_TIME = ENV.get('DEFAULT_SLEEP_TIME', 10)
 
 
+def to_seconds(duration: str) -> int:
+    """Returns a value of a string duration in seconds
+       This string is expected in HH:MM:SS or MM:SS or SS
+
+    :param timestr: String in a duration format like HH:MM:SS
+    :type timestr: str
+    :return: Value in seconds
+    :rtype: int
+    """
+    total_seconds = 0
+    for second in duration.split(':'):
+        try:
+            total_seconds = total_seconds * 60 + int(second)
+        # If there is an error just return 0
+        except ValueError:
+            logger.warn(f"{str} is not a valid duration")
+            return 0
+    return total_seconds
+
+
 def get_request_retry(url: str, auth: AuthBase,
                       params: Dict[str, Union[str, int]]) -> requests.Response:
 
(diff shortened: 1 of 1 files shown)


online_meetings/report.py
```diff
@@ -21,6 +21,26 @@
 DEFAULT_SLEEP_TIME = ENV.get('DEFAULT_SLEEP_TIME', 10)
 
 
+def to_seconds(duration: str) -> int:
+    """Returns a value of a string duration in seconds
+       This string is expected in HH:MM:SS or MM:SS or SS
```
ssciolla, line 26: Have you seen values in all these time formats?

zqian: Right. The API returns duration value in that format

online_meetings/report.py
```diff
@@ -112,6 +132,13 @@ def run_report(api_url: str, json_attribute_name: str,
         total_df.drop_duplicates("uuid", inplace=True)
         logger.info(f"Dataframe with duplicates removed: {len(total_df)}")
 
+    # Create a new calculated column of partipants and duration in the data frame
+    if {'participants', 'duration'}.issubset(total_df.columns):
+        # Fill N/A in participants and duration to zero
+        total_df = total_df.fillna({'participants': 0, 'duration': 0})
+        total_df['particpant_minutes_est'] = total_df.apply(
+            lambda x: round(x['participants'] * to_seconds(x['duration']) / 60, 2), axis=1)
```
zqian, line 140: Do we need to track to the second decimal place?

ssciolla: If it's minutes we're calculating, that makes sense to me. Maybe even more precision makes sense, if we're trying to produce an accurate aggregate number.

online_meetings/report.py
jonespm: Yes, all of the values appear to be in this format. I see some meetings lasting exactly 24 hours but nothing longer. It looks like Zoom cuts off at that and there's nothing beyond.

online_meetings/report.py
jonespm: I added the round just because some of these were repeating decimals. However this calculation is an estimate that's and the fraction isn't super relevant. We could go either way on this just round to the minute or just leave it at X digits.

online_meetings/report.py
```diff
@@ -21,6 +21,26 @@
...
+    :rtype: int
+    """
+    total_seconds = 0
+    for second in duration.split(':'):
+        try:
+            total_seconds = total_seconds * 60 + int(second)
+        # If there is an error just return 0
+        except ValueError:
+            logger.warn(f"{str} is not a valid duration")
+            return 0
```
ssciolla, line 40: This looks like it will work. I was a little confused at first by the variable names, since the value of `total_seconds` doesn't signify seconds until the end. Not sure what would be better.

online_meetings/report.py
```diff
@@ -112,6 +132,13 @@ def run_report(api_url: str, json_attribute_name: str,
         total_df.drop_duplicates("uuid", inplace=True)
         logger.info(f"Dataframe with duplicates removed: {len(total_df)}")
 
+    # Create a new calculated column of partipants and duration in the data frame
```
ssciolla, line 135:
```suggestion
    # Create a new calculated column in the DataFrame using participants and duration
```

online_meetings/report.py
```diff
@@ -21,6 +21,26 @@
...
+    :return: Value in seconds
+    :rtype: int
+    """
+    total_seconds = 0
+    for second in duration.split(':'):
+        try:
+            total_seconds = total_seconds * 60 + int(second)
+        # If there is an error just return 0
+        except ValueError:
+            logger.warn(f"{str} is not a valid duration")
```
ssciolla, line 39: Looks like `warn` is deprecated: https://docs.python.org/3/library/logging.html#logging.Logger.warning

jonespm: Always forget this!

online_meetings/report.py
jonespm: Yeah, it's just used as an accumulator in the for loop.

PR #102, opened 2020-04-16
Add fields that need to be dropped by zoom

### LINKED ISSUE #100: Fields need to be dropped from data
On the 04/15 meeting it was discussed that meeting id and topic need to be dropped from the Zoom (and probably Bluejeans) reports. Also anything containing the meeting id (like the URL) need to be dropped as well.

We need a way of doing this easily per data source. 

This includes from Zoom
```
total_meetings -> id, topic
total_webinars -> id, topic
zoom_courses_meetings -> meeting_number, topic, join_url
```

The Bluejeans script should drop meetingTitle and meetingId.

jonespm:
Refactor some of the csv writing that's redundant out to a new class.
Plan is for this to be used to write SQL instead, but made some sense to
have some of the drop and other logic in here.

Fixes #100

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,6 +82,7 @@ This includes the creation of a configuration file called `env.json`. Complete t
     `CANVAS_URL` | The Canvas instance URL to be used as the base URL for API requests that use the `CANVAS TOKEN`.
     `CANVAS_TOKEN` | The Canvas token used for authenticating to the API when not using the U-M API Directory.
     `NUM_ASYNC_WORKERS` | Number of workers for asynchronous API calls; the default is 8.
+    `DROP_DATABASE_COLUMNS` | This is a special object with the key the table and value as a list of columns to drop. See the sample in the file as an example.
     `DEFAULT_SLEEP_TIME` | Amount of time to sleep between re-tries when given a 429 error and no Retry-After in the response headers.
     `WAREHOUSE_INCREMENT` | A 17-digit number that can be added to some Canvas IDs to create Unizin Data Warehouse IDs (not in use currently).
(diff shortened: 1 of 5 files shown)


jonespm:
Rather than having a list of fields that need to be dropped, Sam and I were talking about having a schema defined that represents the format of the table that needs to be loaded, maybe in JSON format, maybe in Python/ORM. I don't know yet. 

I was [looking at this article](https://www.codementor.io/@bruce3557/graceful-data-ingestion-with-sqlalchemy-and-pandas-pft7ddcy6) and idea #4 looks interesting. Giving a lot more control and structure to the ORM. This would be a large change the this project though. Maybe it would be better to just write something simpler to do this.

lsloan:
So, you're thinking of something like a JSON transform.  I'd started working on that kind of thing with Anthony around 1.5 years ago.  We wanted something to transform Caliper 1.0 format events to 1.1 format.  I was thinking of using something like jq, but within Python.  There are some XML transformation libraries around (some use XSLT), but at the time I couldn't find JSON ones.

We had thought that if Anthony was defining the Caliper 1.1 schema in JSON itself, maybe that could be used as a map.  Maybe a Caliper 1.0 schema maps the event JSON into data structures, then the 1.1 schema maps those structures back to JSON.

Maybe that kind of schema-as-map idea could be useful to you for this.

jonespm:
This will *probably* end up being closed but I'm leaving it open until the work on the database is completed.

jonespm:
I'm just going to close this for now and probably re-work it for the LTI effort

Issue #103, opened 2020-04-17
Use key value instead of row number for pulling canvasdatadate from UDW


ssciolla:
It seems like the test job failed this morning because the key-value pairs in the `unizin_metadata` table in UDW came back switched, leading to the code trying to parse `schemaversion` as a DateTime. We should just avoid this possible issue by using `pandas` to find the right row, rather than hardcoding a row number.

```
ValueError: time data '4.2.6' does not match format '%Y-%m-%d %H:%M:%S.%f%z' (match)
```

zqian:
Verified that the data_updated_at time "UNIZIN_DATA_WAREHOUSE" source has been properly populated for recent runs. 

We may want to surface the data_updated_at values for all data sources in the Tableau view.

PR #104, opened 2020-04-17
Identify and pull canvasdatadate using key name and value column (#103)

ssciolla:
This PR makes a small change that aims to avoid errors by moving from a hardcoded-index-based approach to a name-based approach for identifying and pulling the `canvasdatadate` value from the UDW's `unizin_metadata` table. This PR aims to resolve issue #103.

### DIFF
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -245,8 +245,12 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
     section_df = pd.merge(section_df, sis_section_df, on='canvas_id', how='left')
 
     # Record data source info for UDW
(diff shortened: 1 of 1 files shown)


course_inventory/inventory.py
```diff
@@ -246,7 +246,7 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
 
     # Record data source info for UDW
     udw_meta_df = pd.read_sql('SELECT * FROM unizin_metadata;', udw_conn)
```
zqian, line 248: we may just retrieve the canvasdatadate row only:

select * from unizin_metadata where key='canvasdatadate'

ssciolla: Done!

PR #106, opened 2020-04-17
Move sleep until after MySQL container check (#105)

### LINKED ISSUE #105: move sleep in DB wait loop to after check
The current incremental wait for the DB is better than the original 30-second wait, but even just one 3-second wait when the DB is definitely ready is annoying.  Move the sleep from before the DB check in the loop to just after the check.  (Move the "snail" message, too.)

For example:

```diff
diff --git a/run_jobs.py b/run_jobs.py
index 1e78708..377b020 100644
--- a/run_jobs.py
+++ b/run_jobs.py
@@ -112,11 +112,10 @@ if __name__ == '__main__':

     if how_started == 'DOCKER_COMPOSE':
         # Wait for MySQL container to finish setting up
-        logger.info('Waiting for the MySQL snail')
+        logger.info('Checking whether MySQL is ready')
         # If it's not ready in two minutes, exit
         num_loops = 40
         for i in range(1, num_loops + 1):
-            time.sleep(3.0)
             try:
                 db_creator_obj.set_up()
                 db_creator_obj.tear_down()
@@ -127,7 +126,11 @@ if __name__ == '__main__':
                     logger.error('MySQL was not available')
                     sys.exit(1)
                 else:
-                    logger.debug('Still waiting!')
+                    if (i == 1):
+                        logger.info('Waiting for the MySQL snail')
+                    else:
+                        logger.debug('Still waiting!')
+            time.sleep(3.0)

     # Apply any new migrations
     logger.info('Applying any new migrations')
```

I'm sorry I didn't think of this during the PR for this feature.  It didn't occur to me until I saw the wait in action in my own code after merging from upstream.

ssciolla:
This PR modifies the sleep mechanism for use with `docker-compose` so that the database check happens before the `time.sleep(1)`. The PR aims to resolves #105.

### DIFF
--- a/run_jobs.py
+++ b/run_jobs.py
@@ -111,8 +111,8 @@ if __name__ == '__main__':
     if how_started == 'DOCKER_COMPOSE':
         # Wait for MySQL container to finish setting up
         # If it's not ready in two minutes, exit
-        num_loops = 41
-        for i in range(1, num_loops + 1):
+        num_loops = 40
+        for i in range(num_loops + 1):
             try:
                 db_creator_obj.set_up()
                 db_creator_obj.tear_down()
@@ -123,12 +123,11 @@ if __name__ == '__main__':
                     logger.error('MySQL was not available')
                     sys.exit(1)
                 else:
-                    if i == 1:
+                    if i == 0:
                         logger.info('Waiting for the MySQL snail')
                     else:
                         logger.debug('Still waiting!')
                     time.sleep(3.0)
-
 
     # Apply any new migrations
     logger.info('Applying any new migrations')
--- a/run_jobs.py
+++ b/run_jobs.py
@@ -110,11 +110,9 @@ if __name__ == '__main__':
 
     if how_started == 'DOCKER_COMPOSE':
         # Wait for MySQL container to finish setting up
-        logger.info('Waiting for the MySQL snail')
         # If it's not ready in two minutes, exit
-        num_loops = 40
+        num_loops = 41
         for i in range(1, num_loops + 1):
-            time.sleep(3.0)
             try:
                 db_creator_obj.set_up()
                 db_creator_obj.tear_down()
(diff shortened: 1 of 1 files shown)


run_jobs.py
```diff
@@ -110,11 +110,9 @@ def run_jobs(self) -> None:
 
     if how_started == 'DOCKER_COMPOSE':
         # Wait for MySQL container to finish setting up
-        logger.info('Waiting for the MySQL snail')
         # If it's not ready in two minutes, exit
-        num_loops = 40
+        num_loops = 41
         for i in range(1, num_loops + 1):
```
lsloan, line 115: nit: Rather than increasing the loop count to 41 (which I neglected to account for when I suggested moving the `sleep()` call), why not just let the range start with zero instead of 1…

```suggestion
        num_loops = 40
        for i in range(num_loops + 1):
```

run_jobs.py
```diff
@@ -125,7 +123,12 @@ def run_jobs(self) -> None:
                     logger.error('MySQL was not available')
                     sys.exit(1)
                 else:
-                    logger.debug('Still waiting!')
+                    if i == 1:
```
lsloan, line 126: nit: If the loop range is changed to start with 0 (as suggested in my other comment), then this condition should check for that…

```suggestion
                    if i < 1:
```

or…

```suggestion
                    if not i:
```

(I wanted to suggest changing the range initial value to 0 when I wrote the issue, but I thought it might be seen as nitpicky.  But in this case, using 0 actually lets us keep the original loop count.)

run_jobs.py
ssciolla: Yeah, this is better. Change made.

run_jobs.py
ssciolla: I chose to use `if i == 0`, not to be contrarian, I just think it's a bit more specific? I dunno, it's probably fine any of these ways.

lsloan: I like the `not i` solution, because it seems cool.  There's a slight feeling of risk associated with it, though, isn't there?  The `== 0` solution is fine and risk free, but `< 1` is a character shorter.  (Whee!)

lsloan:
Yeah, looks great.  :shipit:

Issue #107, opened 2020-04-22
Fix Codacy integration


ssciolla:
I’m seeing different dashboards and grades (B versus A) when I click on the Codacy badge in the README versus clicking on the results of a scan of a PR. The URLs are completely different:

https://app.codacy.com/gh/tl-its-umich-edu/course-inventory/dashboard 
https://app.codacy.com/manual/ITS_Teaching_And_Learning/course-inventory/dashboard

I imagine there are slightly different settings in each version, resulting in the different grades? Can we consolidate and choose the best settings?

@jonespm, @lsloan, any insights here?

lsloan:
we should be using the dashboard at (https://www.codacy.com/gh/tl-its-umich-edu/course-inventory) because that's the one associated with our GH organization, (https://github.com/tl-its-umich-edu).  Note that the other one was created with a "manual" process.

Also, I recently requested a free preview of the full Codacy service for the org. associated with that first dashboard.  @zqian recently said she would look into purchasing the full Codacy service for TL.  That purchase should be applied to that org.

ssciolla:
K, so maybe this doesn't require a code change.

jonespm:
Yeah, we'd have to look more at these organizations. Codacy is a little confusing. (to me)

When someone adds a project to Codacy they seem to always be associated with that project. If you turn on Codacy to comment it will comment as that user. I don't know if this has changed or not but it can be confusing.

You can move projects to an organization. I believe "tl-its-umich-edu" is one that was used to trial MWrite. Because that repo is private we have to pay for Codacy to use it. "ITS_Teaching_And_Learning" has all of our public projects.

I haven't paid much attention to the "tl-its-umich-edu".

It looks like they have new Github and Legacy organizations now. I don't know how that works. 

Codacy charges per user per month, and it's about $180/user/year so I think we'd need to think about that as we wouldn't want to include those not using it. It would save a LOT of money moving everything to Open Source. ;)

jonespm:
It's also not entirely clear to me which Codacy "grabs" the Github project if multiple people try to add it. Really IMHO only one person should be managing Codacy at least for the initial addition of projects. I remember working through this with Zhen and getting this disconnected was a ton of steps.

ssciolla:
Thanks for the extra information. FYI, I don't see the `course-inventory` project in the `tl-its-umich-edu` organization:

[...]

PR #108, opened 2020-04-22
Update Codacy badge to use legacy organization (for now) (#107)

ssciolla:
This PR changes the Codacy badge in the README so the organization in use matches the legacy organization in use by our other projects. The PR aims to resolve issue #107.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # course-inventory
 
-[![Codacy Badge](https://api.codacy.com/project/badge/Grade/431390d7c73d47cfbf9654f0182ba6d6)](https://www.codacy.com/gh/tl-its-umich-edu/course-inventory)
+[![Codacy Badge](https://api.codacy.com/project/badge/Grade/299102a0330d4dfd8e9f8d5ef36362b5)](https://www.codacy.com/manual/ITS_Teaching_And_Learning/course-inventory?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=tl-its-umich-edu/course-inventory&amp;utm_campaign=Badge_Grade)
 
 <!-- show table of contents in MacDown, others -->
 \[TOC]


README.md
```diff
@@ -1,6 +1,6 @@
 # course-inventory
 
-[![Codacy Badge](https://api.codacy.com/project/badge/Grade/431390d7c73d47cfbf9654f0182ba6d6)](https://www.codacy.com/gh/tl-its-umich-edu/course-inventory)
+[![Codacy Badge](https://api.codacy.com/project/badge/Grade/299102a0330d4dfd8e9f8d5ef36362b5)](https://www.codacy.com/manual/ITS_Teaching_And_Learning/course-inventory?utm_source=github.com&amp;utm_medium=referral&amp;utm_content=tl-its-umich-edu/course-inventory&amp;utm_campaign=Badge_Grade)
```
pushyamig, line 4: What are all these params you are adding after the Question mark? can you refer to the documentation that made you chose those options?

ssciolla: Sorry, in Codacy, on the repository page under the General tab, there is a place where you can copy the Codacy Badge in Markdown. I got this from there.

ssciolla: I could remove the parameters if you want?

pushyamig: That's what codacy seems to suggest then you shouldn't be removing it.

ssciolla: Based on what @lsloan had before, they probably aren't necessary and are primarily used for Codacy's own analytics. Not entirely sure though.

pushyamig: I am ok either way, Looking at this [article](https://support.codacy.com/hc/en-us/articles/212799365-Badges) they are saying if you want to track code performance for the different branch you can use param for doing that. I don't feel the need to dig deep-down what all those params mean. I got curious about what those are and purpose

README.md
```diff
@@ -1,6 +1,6 @@
...
```
lsloan, line 3: As discussed earlier, I don't think the params on the link URL are necessary, but maybe they're useful for analytics.

Issue #109, opened 2020-04-22
Remove self.conn, set_up, and tear_down methods from DBCreator


ssciolla:
Aside from being confusingly named (since they're only being used now for setting up a connection), the utility functions `set_up` and `tear_down` and the associated instance variable `self.conn` seem unnecessary for managing a transitory connection object. Users of `DBCreator` can just use the SQLAlchemy API from the `self.engine` object if needed, making sure to close any connections they open. See the following example.

```
db_creator_obj = DBCreator(INVENTORY_DB, APPEND_TABLE_NAMES)
conn = db_creator_obj.engine.connect()
conn.close()
```

zqian:
verified the set_up, tear_down and self.conn are no longer in code base.

Issue #84, opened 2020-04-09
make DBCreator class implement fluent interface


lsloan:
For convenience, make the `DBCreator` class implement a [fluent interface](https://en.wikipedia.org/wiki/Fluent_interface).  In its most basic form, this means all the class' methods that currently return `None` would instead return the instance of `DBCreator` (i.e., `self`).

Such an interface would allow code like this:

```python
    db_creator_obj = DBCreator(INVENTORY_DB, APPEND_TABLE_NAMES)
    db_creator_obj.set_up()
    db_creator_obj.drop_records()
    db_creator_obj.tear_down()
```

To be written as this instead:

```python
    db_creator_obj = (DBCreator(INVENTORY_DB, APPEND_TABLE_NAMES)
        .set_up().drop_records().tear_down())
```

ssciolla:
Makes sense. I'll try to handle this with some other changes I've been looking at to this class.

PR #110, opened 2020-04-22
Improve DBCreator (#66, #84, #109)

ssciolla:
This PR makes changes to `DBCreator`, and usages, to resolve a few issues that have accumulated related to the class. The issues and how they were resolved are described below in the task list. The PR aims to resolve issues #66, #84, and #109.

- [x] Add `DEFAULT_APPEND_TABLE_NAMES` to the `db_creator` module, and update references to `APPEND_TABLE_NAMES`, so that the metadata tables don't have to be specified in the config. (issue #66)
- [x] Implement a fluent interface for `DBCreator`, which was requested by @lsloan (issue #84).
- [x] Remove `self.conn`, `self.set_up`, and `self.tear_down` from `DBCreator` class (these just caused confusion and really we can use the SQLAlchemy API off of `self.engine` for creating and dispensing with connection objects as needed) (issue #109).

### DIFF
--- a/db/db_creator.py
+++ b/db/db_creator.py
@@ -1,10 +1,13 @@
+# This is needed for type hinting with fluent interfaces
+from __future__ import annotations
+
 # standard libraries
 import logging, os
-from typing import Dict, Sequence
+from typing import Dict, List, Sequence, Union
 from urllib.parse import quote_plus
 
 # third-party libraries
-from sqlalchemy.engine import create_engine
+from sqlalchemy.engine import create_engine, Connection, Engine
 from yoyo import get_backend, read_migrations
 
 logger = logging.getLogger(__name__)
@@ -12,18 +15,21 @@ logger = logging.getLogger(__name__)
 ROOT_DIR = os.path.dirname(os.path.abspath(__file__))
 MIGRATIONS_PATH = os.path.join(ROOT_DIR, 'migrations')
 
+# The metadata tables are append tables by default.
+DEFAULT_APPEND_TABLES = ['job_run', 'data_source_status']
+
 
 class DBCreator:
 
     def __init__(
         self,
         db_params: Dict[str, str],
-        append_table_names: Sequence[str] = []
+        append_table_names: List[str] = []
     ) -> None:
 
-        self.db_name = db_params['dbname']
-        self.conn = None
-        self.conn_str = (
+        self.db_name: str = db_params['dbname']
+        self.conn: Union[Connection, None] = None
+        self.conn_str: str = (
             'mysql+mysqldb' +
             f"://{db_params['user']}" +
             f":{quote_plus(db_params['password'])}" +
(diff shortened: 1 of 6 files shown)


db/db_creator.py
```diff
@@ -1,70 +1,73 @@
...
+    def drop_records(self) -> DBCreator:
         logger.debug('drop_records')
-        self.conn.execute('SET FOREIGN_KEY_CHECKS=0;')
+        conn = self.engine.connect()
+        conn.execute('SET FOREIGN_KEY_CHECKS=0;')
         for table_name in self.get_table_names():
             if 'yoyo' not in table_name and table_name not in self.append_table_names:
                 logger.debug(f'Table Name: {table_name}')
-                self.conn.execute(f'DELETE FROM {table_name};')
+                conn.execute(f'DELETE FROM {table_name};')
```
ssciolla, line 66: Codacy is flagging this as a SQL Injection vulnerability, which is fair. Based on the research I've done, using parameters here isn't possible (https://stackoverflow.com/questions/40249590/sqlalchemy-error-when-adding-parameter-to-string-sql-query), and MyLA is doing [roughly the same thing in `cron.py`](https://github.com/tl-its-umich-edu/my-learning-analytics/blob/36a05c27eee83dd702d65666586ebebf0ab4cf0c/dashboard/cron.py#L78-L83).

I think the best way to fix this is to look into switching to an ORM, probably Django or SQLAlchemy, which I could look into as a separate issue.

db/db_creator.py
```diff
@@ -1,70 +1,73 @@
...
             f"@{db_params['host']}" +
             f":{db_params['port']}" +
             f"/{db_params['dbname']}?charset=utf8&ssl=true"
         )
-        self.engine = create_engine(self.conn_str)
-        self.append_table_names = append_table_names
-
-    def set_up(self) -> None:
-        logger.debug('set_up')
-        self.conn = self.engine.connect()
```
lsloan, line 39: I suspect removing `connect()` from here will break the MiVideo code.

db/db_creator.py
lsloan: Aren't there modules that can check for and eliminate SQL injection?  It really doesn't seem like such a big deal, since the name comes from a query of the DB itself.

db/db_creator.py
```diff
@@ -1,70 +1,73 @@
...
-        self.conn.execute('SET FOREIGN_KEY_CHECKS=1;')
+        conn.execute('SET FOREIGN_KEY_CHECKS=1;')
+        return self
 
-    def set_up_database(self) -> None:
-        self.set_up()
-        self.drop_records()
-        self.migrate()
-        self.tear_down()
+    def set_up_database(self) -> DBCreator:
```
lsloan, line 71: Nice job with the addition of the fluent interface.

db/db_creator.py
ssciolla: Yeah, I don't think it's a big deal either right now. [mysqlclient, and by extension MySQLdb](https://mysqlclient.readthedocs.io/) might have a sanitizing function; I haven't looked there specifically. I think I saw somewhere that Postgres (`psycopg2`) has one.

db/db_creator.py
ssciolla: Nope, it will not. You're connecting via `self.engine`, which is how I'm recommending people do this in the future.

ssciolla: Note that I did change a couple lines of `mivideo/extract.py`.

lsloan: Yes.  I just didn't try running a copy of this myself to see whether this change would cause a problem.

ssciolla:
Thanks, @lsloan. I'm going to merge this now and resolve conflicts in PR #115.

PR #112, opened 2020-04-23
fix MiVideo `data_updated_at` timestamp (#111)

### LINKED ISSUE #111: MiVideo `data_updated_at` timestamp is wrong
I'm consistently seeing dates of `1970-01-01 00:00:02` in the `data_source_status` records associated with Big Query generated by the `mivideo` job.

https://github.com/tl-its-umich-edu/course-inventory/blob/fe54c54608344b77425f6dbcf4c1d36bf84eda05/mivideo/mivideo_extract.py#L126-L129

Based on my brief testing, if we change line 128 above to `'data_updated_at': pd.to_datetime(time.time(), unit='s', utc=True)`, this should be fixed.

lsloan:
Fixes #111.

Specify the units of the source timestamp so Pandas will convert it to a string representation correctly.

### DIFF
--- a/mivideo/mivideo_extract.py
+++ b/mivideo/mivideo_extract.py
@@ -125,7 +125,7 @@ class MiVideoExtract(object):
 
         return [{
             'data_source_name': ValidDataSourceName.UNIZIN_DATA_PLATFORM_EVENTS,
-            'data_updated_at': pd.to_datetime(time.time(), unit='s', utc=True)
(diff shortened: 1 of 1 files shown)


lsloan:
Apologies for the multiple commits.  I'd committed with the wrong key, an unverified one, when my usual key had expired.  I renewed my verified key and committed again.

ssciolla:
LGTM! Might want to squash this one.

lsloan:
Yes, **_this_** is a good reason to squash.

Issue #113, opened 2020-04-23
Accomodate multiple terms in database


ssciolla:
As the Winter term ends, we need to figure out how to handle multiple terms in the database. This will likely involve adding a `term` table to the database schema, as well as fetching some basic information about terms from Canvas. Here are some other questions raised during previous discussion.

* Spring/Summer actually is three different terms, so the `course_inventory` job will need to be modified to fetch data for multiple terms.
* We would probably need a way to only drop records associated with a specific term, which would require a more sophisticated `drop_records` method. (This idea needs some more discussion and thinking, so I'm moving it to a separate issue.)

pushyamig:
I was able to test this from master branch (05/05/20 ) and I was able to run all 3 processes ["COURSE_INVENTORY", "MIVIDEO", "CANVAS_ZOOM_MEETINGS"]. Look good!

PR #115, opened 2020-04-27
Add term table, accomodate multiple terms, fix course.sis_id bug (#113, #117)

ssciolla:
This PR modifies the `COURSE_INVENTORY`job and the database schema as to be able to collect and store data related to multiple academic terms. Some additional details are provided in the task list below. This PR aims to resolve issues #113 and #117.

- [x] Add migration with new `term` table and with a `term_id` column and foreign key constraint to the `course` table.
- [x] Make `CANVAS_TERM_ID` configuration variable plural, i.e. `CANVAS_TERM_IDS`.
- [x] Add function that pulls data for terms specified in the config.
- [x] Modify `get_course_data_from_api` function to accept multiple terms, nest fetching process inside`for` loop, and then combine all results into one `course_df` (ensuring that future processes will consider courses from the multiple terms specified).
- [x] Write contents of `term_df` to database.
- [x] Add migration changing the data type of `course.sis_id` to `VARCHAR(15)` and type coercion to string of those values (see issue #117 for details).
- [x] Update `online_meetings/canvas_zoom_meetings.py` to handle multiple terms.
- [x] Update README.

### DIFF
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -135,39 +135,44 @@ def slim_down_course_data(course_data: List[Dict]) -> List[Dict]:
     return slim_course_dicts
 
 
-def gather_course_data_from_api(account_id: int, term_id: int) -> pd.DataFrame:
+def gather_course_data_from_api(account_id: int, term_ids: Sequence[int]) -> pd.DataFrame:
     logger.info('** gather_course_data_from_api')
     url_ending_with_scope = f'{API_SCOPE_PREFIX}/accounts/{account_id}/courses'
-    params = {
-        'with_enrollments': True,
-        'enrollment_type': ['student', 'teacher'],
-        'enrollment_term_id': term_id,
-        'per_page': 100,
-        'include': ['total_students']
-    }
 
-    # Make first course request
-    page_num = 1
-    logger.info(f'Course Page Number: {page_num}')
-    response = make_request_using_api_utils(url_ending_with_scope, params)
-    all_course_data = json.loads(response.text)
-    course_dicts = slim_down_course_data(all_course_data)
-    more_pages = True
+    course_dicts = []
+    for term_id in term_ids:
+        logger.info(f'Processing term {term_id}')
 
-    while more_pages:
-        next_params = API_UTIL.get_next_page(response)
-        if next_params:
-            page_num += 1
-            logger.info(f'Course Page Number: {page_num}')
-            response = make_request_using_api_utils(url_ending_with_scope, next_params)
-            all_course_data = json.loads(response.text)
-            course_dicts += slim_down_course_data(all_course_data)
-        else:
-            logger.info('No more pages!')
-            more_pages = False
+        params = {
+            'with_enrollments': True,
+            'enrollment_type': ['student', 'teacher'],
+            'enrollment_term_id': term_id,
+            'per_page': 100,
+            'include': ['total_students']
+        }
+
+        # Make first course request
+        page_num = 1
+        logger.info(f'Course Page Number: {page_num}')
+        response = make_request_using_api_utils(url_ending_with_scope, params)
+        all_course_data = json.loads(response.text)
+        course_dicts += slim_down_course_data(all_course_data)
+        more_pages = True
+
+        while more_pages:
+            next_params = API_UTIL.get_next_page(response)
+            if next_params:
+                page_num += 1
+                logger.info(f'Course Page Number: {page_num}')
+                response = make_request_using_api_utils(url_ending_with_scope, next_params)
+                all_course_data = json.loads(response.text)
+                course_dicts += slim_down_course_data(all_course_data)
+            else:
+                logger.info('No more pages!')
+                more_pages = False
 
     num_course_dicts = len(course_dicts)
-    logger.info(f'Total course records: {num_course_dicts}')
+    logger.info(f'Total course records for all active terms: {num_course_dicts}')
     course_dicts_with_students = []
     for course_dict in course_dicts:
         if course_dict['total_students'] > 0:
@@ -243,7 +248,7 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
     new_term_df = gather_new_term_data_from_api(ACCOUNT_ID, TERM_IDS, db_creator_obj)
 
     # Gather course data
-    course_df = gather_course_data_from_api(ACCOUNT_ID, TERM_IDS[0])
+    course_df = gather_course_data_from_api(ACCOUNT_ID, TERM_IDS)
 
     logger.info("*** Fetching the published date ***")
     course_available_df = course_df.loc[course_df.workflow_state == 'available'].copy()
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -67,55 +67,38 @@ def make_request_using_api_utils(url: str, params: Dict[str, Any] = {}) -> Respo
     return response
 
 
-def gather_new_term_data_from_api(
-    account_id: int,
-    term_ids: Sequence[int],
-    creator_obj: DBCreator
-) -> pd.DataFrame:
-
+def gather_new_term_data_from_api(account_id: int, term_ids: Sequence[int]) -> pd.DataFrame:
     logger.info('** gather_new_term_data_from_api')
 
-    # Find terms that don't already have records
-    term_df = pd.read_sql('term', creator_obj.engine)
-    new_term_ids = [
-        term_id for term_id in term_ids
-        if term_id not in term_df['canvas_id'].to_list()
-    ]
+    # Fetch data for terms from config
+    logger.info(f'Canvas terms specified in config: {term_ids}')
+    url_ending_with_scope = f'{API_SCOPE_PREFIX}/accounts/{account_id}/terms/'
 
-    if len(new_term_ids) == 0:
-        logger.info('No new term IDs were found; returning empty DataFrame')
-        return pd.DataFrame({})
-    else:
-        logger.info(f'Found new term ID(s) in config: {new_term_ids}')
+    term_dicts = []
+    for term_id in term_ids:
+        logger.info(f'Pulling data for term number {term_id}')
+        term_url_ending = url_ending_with_scope + str(term_id)
+        response = make_request_using_api_utils(term_url_ending)
 
-        # Fetch data for new terms
-        url_ending_with_scope = f'{API_SCOPE_PREFIX}/accounts/{account_id}/terms/'
+        term_data = json.loads(response.text)
+        slim_term_dict = {
+            'canvas_id': term_data['id'],
+            'name': term_data['name'],
+            'sis_id': int(term_data['sis_term_id']),
+            'start_at': pd.to_datetime(
+                term_data['start_at'],
+                format=CANVAS_DATETIME_FORMAT
+            ),
+            'end_at': pd.to_datetime(
+                term_data['end_at'],
+                format=CANVAS_DATETIME_FORMAT
+            )
+        }
+        term_dicts.append(slim_term_dict)
 
-        new_term_dicts = []
-        for new_term_id in new_term_ids:
-            logger.info(f'Pulling data for Canvas term number {new_term_id}')
-            term_url_ending = url_ending_with_scope + str(new_term_id)
-            response = make_request_using_api_utils(term_url_ending)
-
-            term_dict = json.loads(response.text)
-            new_slim_term_dict = {
-                'canvas_id': term_dict['id'],
-                'name': term_dict['name'],
-                'sis_id': int(term_dict['sis_term_id']),
-                'start_at': pd.to_datetime(
-                    term_dict['start_at'],
-                    format=CANVAS_DATETIME_FORMAT
-                ),
-                'end_at': pd.to_datetime(
-                    term_dict['end_at'],
-                    format=CANVAS_DATETIME_FORMAT
-                )
-            }
-            new_term_dicts.append(new_slim_term_dict)
-
-        new_term_df = pd.DataFrame(new_term_dicts)
-        logger.debug(new_term_df.head())
-        return new_term_df
+    term_df = pd.DataFrame(term_dicts)
+    logger.debug(term_df.head())
+    return term_df
 
 
 def slim_down_course_data(course_data: List[Dict]) -> List[Dict]:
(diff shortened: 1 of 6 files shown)


course_inventory/inventory.py
```diff
@@ -14,18 +14,18 @@
...
+from course_inventory.published_date import FetchPublishedDate
 
 
 # Initialize settings and globals
 
 logger = logging.getLogger(__name__)
 
 ACCOUNT_ID = ENV.get('CANVAS_ACCOUNT_ID', 1)
-TERM_ID = ENV['CANVAS_TERM_ID']
+TERM_IDS = ENV['CANVAS_TERM_IDS']
```
lsloan, line 28: What verifies that `CANVAS_TERM_IDS` is definitely an array from the JSON and contains values?

course_inventory/inventory.py
```diff
@@ -65,6 +67,57 @@ def make_request_using_api_utils(url: str, params: Dict[str, Any] = {}) -> Respo
     return response
 
 
+def gather_new_term_data_from_api(
+    account_id: int,
+    term_ids: Sequence[int],
+    creator_obj: DBCreator
+) -> pd.DataFrame:
```
lsloan, line 74: New style of putting each function argument on a separate line?

course_inventory/inventory.py
```diff
@@ -14,18 +14,18 @@
...
 from environ import ENV
 from vocab import ValidDataSourceName
-from .async_enroll_gatherer import AsyncEnrollGatherer
-from .canvas_course_usage import CanvasCourseUsage
-from .gql_queries import queries as QUERIES
-from .published_date import FetchPublishedDate
+from course_inventory.async_enroll_gatherer import AsyncEnrollGatherer
+from course_inventory.canvas_course_usage import CanvasCourseUsage
+from course_inventory.gql_queries import queries as QUERIES
+from course_inventory.published_date import FetchPublishedDate
```
lsloan, line 20: Was Codacy complaining about the lack of `course_inventory` here?

course_inventory/inventory.py
```diff
@@ -39,6 +39,8 @@
 INVENTORY_DB = ENV['INVENTORY_DB']
 APPEND_TABLE_NAMES = ENV.get('APPEND_TABLE_NAMES', ['job_run', 'data_source_status'])
 
+CANVAS_DATETIME_FORMAT = '%Y-%m-%dT%H:%M:%SZ'
```
lsloan, line 42: Isn't this format (ISO 8601 with `Z` shorthand for `+00:00`) already a constant in `datetime` or some similar module?

course_inventory/inventory.py
ssciolla: If it gets too long, why not?

course_inventory/inventory.py
ssciolla: I guess nothing. This is a key configuration variable, as the application needs this input to direct what data to gather. I don't think there is a good default. What are you suggesting here?

course_inventory/inventory.py
ssciolla: I don't think it was? From what I've seen online lately, absolute imports seem preferred over relative ones. I saw that you were doing it in `mivideo` with `queries`, so I figured I'd make the change.

course_inventory/inventory.py
ssciolla: It does seem to basically be the ISO Format. I didn't see this defined as a constant in `datetime`. Did you see it somewhere? This was hardcoded elsewhere in this file, I just extracted it to be a global variable since I needed it to.

course_inventory/inventory.py
```diff
@@ -268,6 +336,11 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
 
     if CREATE_CSVS:
         # Generate CSV Output
+        if len(new_term_df) > 0:
```
zqian, line 339: The term.csv file only contains new term information -- is this as desired?

db/migrations/0014.change_course_sis_id_data_type.py
```diff
@@ -0,0 +1,10 @@
+#
+# file: migrations/0014.change_course_sis_id_data_type.py
```
zqian, line 2: Is this change belongs to this issue? I thought the sis_id change was already merged in?

course_inventory/inventory.py
```diff
@@ -183,10 +242,17 @@ def pull_sis_section_data_from_udw(section_ids: Sequence[int], conn: connection)
...
+    db_creator_obj = DBCreator(INVENTORY_DB, APPEND_TABLE_NAMES)
+
     logger.info('Making requests against the Canvas API')
 
+    # Gather new_term_data
+    new_term_df = gather_new_term_data_from_api(ACCOUNT_ID, TERM_IDS, db_creator_obj)
+
     # Gather course data
-    course_df = gather_course_data_from_api(ACCOUNT_ID, TERM_ID)
+    course_df = gather_course_data_from_api(ACCOUNT_ID, TERM_IDS)
```
zqian, line 255: instead of changing the gather_course_data_from_api() function, how about adding the term loop here, and just keep the gather_course_data_from_api() definition the same as before?

course_inventory/inventory.py
ssciolla: You tell me; I can change it if you want. It made sense to me to use the `append_table` logic here; we're not using an append table with CSV generation elsewhere yet. I could also just not include this in the CSV generation.

db/migrations/0014.change_course_sis_id_data_type.py
ssciolla: This is a new issue, which I've documented here #117. I was thinking about opening a PR separately, but if anyone tests this PR on the spring/summer terms, it's going to break without this change.

ssciolla: I guess I didn't see the harm in including it if it was documented separately.

course_inventory/inventory.py
ssciolla: I thought about this a little bit. I could do it that way, but I guess I thought we need to combine all the results anyway and this way the filtering could be done all at once. If you'd like me to change it to the way you're suggesting, that's fine, please ask me to do so directly :)

course_inventory/inventory.py
ssciolla: I’ll rethink this in the morning. It may make more sense to not treat this as an append table.

course_inventory/inventory.py
lsloan: I agree.  I asked because there are other function definitions you changed in this PR that could also have their arguments arranged on separate lines, but you didn't do that.  E.g.,…

* https://github.com/tl-its-umich-edu/course-inventory/pull/115/files#diff-3ab4d3be8a800981ae3e09e15194bc0dR142
* https://github.com/tl-its-umich-edu/course-inventory/pull/115/files#diff-9e1a06acd700cfb38555dc5980f1cdb0R163

I know that, "A foolish consistency is the hobgoblin of little minds".  I hope it's not foolish to say this style stands out (in a good way) and I wondered whether it should apply in these other instances, too.

course_inventory/inventory.py
lsloan: OK.  I happened to do it that way because my IDE's autocomplete feature filled that in for me.  I think I'd also read that absolute references are preferred.

course_inventory/inventory.py
lsloan: I suggest that the type of `ENV['CANVAS_TERM_IDS']` is verified to be a list.  (Data coming from the JSON file should be immutable, so maybe it should be a tuple instead.)  It should not be an empty list and its values should not be `None`.

course_inventory/inventory.py
lsloan: I've worked with ISO 8601 many times in Python and PHP, but I can't remember which language had this format assigned to a constant or a shortcut format for it.  I'll check around.  I remember that there was something that used a single character format, like `%c`, as a shortcut for `%Y-%m-%dT%H:%M:%SZ` or `%Y-%m-%dT%H:%M:%S%z`.

db/migrations/0014.change_course_sis_id_data_type.py
lsloan: I think it's fine to include the resolution of #117 in this PR.  It's a small change and it fits closely with the other changes in this PR.  This is not a case where a lot of code for disparate issues is being forced into the same PR.

#117 is mentioned in the PR description and has been added to linked issues.  I think the one thing missing is to include it in the title of the PR.  The title mentions only #113, which might have confused @zqian.  Links to issues in PR titles should be complete or not done at all.

ssciolla: Done!

course_inventory/inventory.py
ssciolla: Okay, I changed the latter one. The former has only two parameters and isn't nearly as long, so I'm going to pass on that one, it that's acceptable.

course_inventory/inventory.py
ssciolla: This is good, I think I lack experience in properly handling inputs. Can you tell me why it shouldn't be an empty array?

lsloan: The code needs at least one term to work with, doesn't it?  Have I misunderstood what it's doing with the term IDs?

ssciolla: Yeah, I think you're right that it would probably break somewhere with no terms. I'm asking in part because of the `canvas_zoom_placements`, which can accept a null value and instead process specific course IDs. I'll think about this some more and commit a possible solution.

jonespm: You could check it, but if you don't you'll just get an exception somewhere, either a KeyError here or some other exception somewhere else. You could use `TERM_IDS = ENV.get("CANVAS_TERM_IDS", [])` here to avoid the KeyError but then you'd still have to 

```
if not isinstance(TERM_IDS, list):
   # Print some error message

```

I don't know *really* how important this is over just letting this throw an exception if this is incorrectly defined since it's just us using it though.

I also think if this became an issue we'd go down defining this as a Json Schema and validating that instead of doing this one off stuff. https://github.com/Julian/jsonschema.

ssciolla: Yeah, I was kind of thinking along those lines, particularly because @lsloan could have brought up this issue with many of the variables in use here. As someone who has used schemas before with XML/Relax NG, I really like the idea of using JSON Schema to validate config files. Seems like it would be a useful tool for collaboration, also.

lsloan:
@ssciolla, any idea why Codacy claims `zoom_course_report` is using "`%s` as argument"?  (Which Codacy can't even display correctly!)

Can we use comments to tell Codacy (or in this case, PyLint specifically) to ignore this error for this section of code? (https://github.com/tl-its-umich-edu/course-inventory/pull/115/files#diff-9e1a06acd700cfb38555dc5980f1cdb0R163)

ssciolla:
> @ssciolla, any idea why Codacy claims `zoom_course_report` is using "`%s` as argument"? (Which Codacy can't even display correctly!)
> 
> Can we use comments to tell Codacy (or in this case, PyLint specifically) to ignore this error for this section of code? (https://github.com/tl-its-umich-edu/course-inventory/pull/115/files#diff-9e1a06acd700cfb38555dc5980f1cdb0R163)

I think that's an issue that was there before; it doesn't like the default mutable argument of `[]`. I'm not really sure I get why it's an issue; if we agree, maybe it's something we can configure in Codacy?

course_inventory/inventory.py
ssciolla: @zqian, after doing some thinking, I decided it was simpler and less confusing to drop this table as well. This should also make the `term_df` output more regular. See [this commit](https://github.com/tl-its-umich-edu/course-inventory/pull/115/commits/32500297adc106142db3cffbdee96a2edbaa7e14).

I'm also thinking that this will simplify what I'm currrently thinking of as the solution for #118 (a `DELETE` with `JOIN`s and a `WHERE`, see issue for SQL).

course_inventory/inventory.py
ssciolla: Also, I changed the term function a bit, so it doesn't need the `creator_obj` anymore, meaning I think I can push this back to one line.

ssciolla: Resolving this for now.

course_inventory/inventory.py
ssciolla: @lsloan, do you think this needs to be resolved now, or can I create a separate issue, referencing the JSON Schema discussion?

lsloan: I don't think it's urgent enough to delay this PR.  Make it an issue.

ssciolla: Thanks, I've created issue #119.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -160,16 +160,30 @@ def get_zoom_course(self, course: canvasapi.course.Course) -> None:
                 self.get_zoom_details(posturl, formdata, course.id)
         return None
 
-    def zoom_course_report(self, canvas_account: int = 1, enrollment_term_id: int = 0,
-                           published: bool = True, add_course_ids: list = None) -> None:
+    def zoom_course_report(
+        self,
+        canvas_account: int = 1,
+        enrollment_term_ids: Sequence[int] = [],
```
jonespm, line 166: You should avoid using a mutable default value as an argument. This should follow this pattern. https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

jonespm: This was mentioned in the Codacy review as an issue as well.

ssciolla: Thank you for flagging this and sharing the link, I'm happy to have read and learned about the anti-pattern. I think I've resolved it here, but there are two other occurrences of this that I introduced to the codebase. I'll open a separate PR to address those.

lsloan: Ah, that makes sense.  Codacy (PyLint) referred to it as a problem with `%s`.  I was distracted by that weirdness and I didn't think of the real potential problem.  I think I would automatically code it with the sentinel value and I've never thought of what could be the problem with specifying the empty list as default argument.

jonespm: Yeah, I think it's a "pretty cool feature" until you run into it in actual code accidentally. ;) Shouldn't use empty lists or dicts as default args.

course_inventory/inventory.py
ssciolla: Gonna go with this approach for now, we can always revisit.

course_inventory/inventory.py
ssciolla: Not gonna worry about this for now.

Issue #58, opened 2020-04-03
Change online_meetings to write output CSV files to the root data directory


ssciolla:
The rest of the CSV output generated by this project writes to the `data` directory, which when using `docker-compose` is mapped to `~/data/course-inventory`. Once the job manager is finished (issue #33), we may want to use `docker-compose` to generate CSVs while continuing to develop the `online_meetings` package.

jonespm:
Yeah we are going to need a storage volume for the Bluejeans, I figure we'll make one in general to dump the csvs. I think we can use oc rsync to semi-easily pull the data down locally without having to add anything special to the container. Though we might be able to install and use this? https://github.com/box/boxcli

ssciolla:
That data ultimately should go to the database, so I'm not sure creating a volume is super high-priority in the OpenShift environment. Locally though, until we get the database stuff set up, you'll want to access to those files.

ssciolla:
I guess depends on what you think the timeline is for database integration and how urgent things need to be automated.

pushyamig:
The energy/time you put in volume container and another temp automated process you can do the DB integration.  Maybe I am missing something? Why is it taking long for BlueJeans DB integration?

jonespm:
BlueJeans will always need a volume because the process has to download and unzip a file. #55 

It's possible that this may be able to be done in memory but that would require more code and also require that there is enough memory to manipulate the zip file.

Maybe there's something similar to this but in a package?
https://stackoverflow.com/a/10909949/3708872

ssciolla:
We can use the `tmp` directory if we need a place to process a zip file, right? Unless the file needs to be available later.

pushyamig:
How much data we are talking about? as it sounds to me like it huge to
unpack the file? Can we increase our container processing power? ( Not sure
how that will apply to local runs)

jonespm:
I don't know how big the file is yet, I haven't tried. It's only in the 10s of MB's size tops though. So I *think* it would fit in memory if we wrote a process to extract it into memory. Most of the code extract to disk that I've seen. I'm hoping there'd be a package already out there that does ZipFile in memory rather than us relying on outdated Stackoverflow code.

pushyamig:
Maybe we need to come up with smart coding that doesn't load all the files
at one time rather like stream approach.

jonespm:
[...]

PR #124, opened 2020-04-28
Incorporate canvas_zoom_meetings into run_jobs to make it able to run

### LINKED ISSUE #122: Incorporate canvas_zoom_meetings into run_jobs to make easier run from Docker
Currently canvas_zoom_meetings is run standalone as a module. It should be invoked from run_jobs instead.

jonespm:
from Docker

Fixes #122


online_meetings/canvas_zoom_meetings.py
```diff
@@ -168,12 +150,12 @@ def zoom_course_report(
         add_course_ids: list = None
```
ssciolla, line 150: I probably should have updated this in my PR. Could we make the type hint here `Union[List[int], None]`?

ssciolla: Or `Sequence` is probably fine here.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -199,24 +181,30 @@ def zoom_course_report(
...
+    zoom_courses_df = pd.DataFrame(zoom_placements.zoom_courses)
+    zoom_courses_df.index.name = "id"
+    zoom_courses_meetings_df = pd.DataFrame(zoom_placements.zoom_courses_meetings)
+    zoom_courses_meetings_df.index.name = "id"
 
-zoom_courses_df = pd.DataFrame(zoom_placements.zoom_courses)
-zoom_courses_df.index.name = "id"
-zoom_courses_meetings_df = pd.DataFrame(zoom_placements.zoom_courses_meetings)
-zoom_courses_meetings_df.index.name = "id"
+    zoom_courses_df.to_csv("zoom_courses.csv")
```
ssciolla, line 201: If you're already changing this line, it might be an appropriate time to just change it to `os.path.join('data', 'zoom_courses.csv')` and resolve issue #58.

ssciolla: Same with the other CSV below.

vocab.py
```diff
@@ -12,8 +12,8 @@ class ValidJobName(Enum):
     
     COURSE_INVENTORY = 'course_inventory.inventory.run_course_inventory'
     MIVIDEO = 'mivideo.mivideo_extract.main'
-    # ONLINE_MEETINGS = 'online_meetings.report...'
-    # ZOOM = 'online_meetings.canvas_zoom_meetings...'
+    CANVAS_ZOOM_MEETINGS = 'online_meetings.canvas_zoom_meetings.main'
+    # ZOOM = 'online_meetings.report...'
```
ssciolla, line 16: Nitpick: Perhaps we could delete this comment now?

vocab.py
```diff
@@ -25,3 +25,4 @@ class ValidDataSourceName(Enum):
     CANVAS_API = auto()
     UNIZIN_DATA_PLATFORM_EVENTS = auto()
     UNIZIN_DATA_WAREHOUSE = auto()
+    CANVAS_ZOOM_MEETINGS = auto()
```
ssciolla, line 28: I think it's a little confusing that this is the same as the `ValidJobName`. The data is coming from the Canvas API and the user interface, correct? I know that might change, but I'd probably create two `data_source_status` dictionaries, one for `CANVAS_API` and `CANVAS_UI`. What do you think?

online_meetings/canvas_zoom_meetings.py
jonespm: Sequence[int] has no attribute remove. Needs to be a List it looks like but I can update it.

online_meetings/canvas_zoom_meetings.py
jonespm: Have you tested that actually works without using the ROOT_DIR? Seems like it would be a good idea though and something that would have to happen for this to work. I think this may need to be something defined in environ.py.

ssciolla: Oh yeah, didn't think about that. Except I am doing this in `inventory`, just specifying `os.path.join('data', 'term.csv')`, for example. Hm...

vocab.py
jonespm: Yeah, the source of the data is both CANVAS_API and from the ZOOM_UI (or maybe ZOOM_LTI). I feel like it's more just ZOOM in this case really. Maybe it could be both of these though. I don't 100% understand how this is represented in the data model as I haven't looked at that yet.

online_meetings/canvas_zoom_meetings.py
ssciolla: Might be better to actually just import `ROOT_DIR` from `environ.py` and build it yourself, like `os.path.join(ROOT_DIR, 'data', 'term.csv')`.

jonespm: Yeah, I could do that if you think that's better over putting it in environ.py. I just don't *think* it would actually write it to the correct location when running from inventory without that ROOT_DIR in there.

ssciolla: I tested it, it is working, I believe because the current working directory is where `run_jobs.py` is. Still, it's likely better to use `ROOT_DIR` so that it will also work when you run it as a standalone application from any directory. If you wanted to make a `DATA_DIR` global in `environ.py`, I think that would be fine, too.

vocab.py
ssciolla: I'll leave it up to you, but the data source(s) should document everywhere data is coming from. I get that it's more confusing with LTI, so whatever works for you.

online_meetings/canvas_zoom_meetings.py
ssciolla: Thinking about it now, not even sure if `inventory.py` can be run as a standalone application anymore; keep getting module import errors...

jonespm: Yeah, I'm trying to think about this. When it's running locally it's convenient to just have the config files in the same directory as the app but it would be *better* if these directories were outside of /app if only so files they're never possibly copied an container we release anywhere.

Historically these would go into either /mnt or /media. Maybe just have it /media/config and /media/data?

I think we could make that more explicit without breaking anything that currently exists out there.

ssciolla: Doesn't `.dockerignore` take care of ignoring certain file types for image builds? I guess I don't quite understand what you're saying.

jonespm: Maybe I'll move that refactor off to a future issue and just deal with the data directory as is for now.

jonespm: Yeah, I guess you're right. I feel like there was always a security risk of storing any files in the same path as your web app because it's possible to navigate to those files and the web server would serve it up as plain text. Though since this isn't a web app it's probably not exploitable, just feels like best practice.

ssciolla: Yeah, sounds like something good to think about in a separate issue. It'd be nice to think through this for other applications, too.

ssciolla: Change looks good, resolving...

PR #126, opened 2020-04-29
Clean up local flake8 file

### LINKED ISSUE #125: Clean up local flake8 (and any other lint config) files
I noticed a few subdirectories had some local configs, I'd like to clean these up and just use the top level one if possible. It looks like the only thing that's failing for mivideo is the line length which we are overriding at the top level.

jonespm:
Fixes #125

### DIFF
--- a/mivideo/.flake8
+++ b/mivideo/.flake8
@@ -1,7 +1,0 @@
-[flake8]
-ignore =
-    E121,E123,E126,E24,E704,W503,W504,  # these are ignored by default
-    #E501,  # line too long
-    E401,  # multiple imports
-    #E111,  # Indentation is not a multiple of 4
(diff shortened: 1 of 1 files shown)


jonespm:
This local config is taking precedence for mivideo over the global. They have a slightly different config and with this file there there are a number of problems with mivideo_extract.py about line length.

pushyamig:
This looks fine to me, If this is resolved we can merge this and have it in the upcoming release. Strictly not needed since it is effected only in local development, but good to have an empty PR's tab

jonespm:
Yeah, I was mostly just waiting on a comment from @lsloan since these were mostly files he created and it passes the 120 line limit and all top level linters but not the default 79.

Issue #72, opened 2020-04-07
Consider re-factoring drop_records in db_creator


ssciolla:
Right now, `DBCreator.drop_records` drops records from all tables not specified in `APPEND_TABLE_NAMES` (also excluding `yoyo`-related tables) and it is only being used in `course_inventory/inventory.py`. However, if other jobs want to use `drop_records`, they will also have to drop tables related to 'inventory.py`, and `inventory.py` will drop their records, and back and forth.

Perhaps `drop_records` should take an optional argument specifying the tables to be dropped.

ssciolla:
I've been thinking about this more and the `APPEND_TABLE_NAMES` idea. I'm kind of wondering now if specifying append tables was the wrong approach; shouldn't the default behavior be NOT to drop records? Instead, maybe we should change `drop_records` to accept a list of table names to drop, which are then validated using `DBCreator.get_table_names()` before records are actually deleted? Then anyone who wants to use this can manage their own tables, specifying tables as part of their job rather than in a configuration file. In my experience, `APPEND_TABLE_NAMES` has proven error-prone and confusing. I guess there was a lack of foresight on my part : (

lsloan:
:shipit:

PR #129, opened 2020-04-30
Refactor drop_records, remove APPEND_TABLE_NAMES (#72)

ssciolla:
This PR primarily makes changes to the `drop_records` method of `DBCreator`, allowing for tables to specified as an additional optional parameter. This change allows us to dispense with the `APPEND_TABLE_NAMES` configuration variable, which I found to be confusing and error-prone. Additional details about this PR will be provided in the task list below. This PR aims to resolve issue #72.

- [x] Refactor `drop_records` to accept an additional optional parameter with specific table names. Table name strings are checked to ensure they are valid table names (i.e. they are accessible through `self.engine.table_names`). If no specific table names are provided, all application-managed (non-yoyo) tables will have their records dropped.
- [x] Add basic docstrings to `DBCreator`.
- [x] Remove all usages of `APPEND_TABLE_NAMES` in `run_jobs.py`, `course_inventory/inventory.py`, and `mivideo/extract.py`.
- [x] Update usage of `DBCreator` in `inventory.py` to specify table names of Canvas data tables.
- [x] Remove mentions of `APPEND_TABLE_NAMES` in `env_blank.json` and `README.md`.


db/db_creator.py
```diff
@@ -41,33 +41,60 @@ def __init__(
...
+        Drops records from either the specified database tables or all of the
+        application-managed database tables; tables managed by yoyo-migrations are ignored.
+        '''
         logger.debug('drop_records')
+        app_table_names = [
+            table_name for table_name in self.get_table_names() if 'yoyo' not in table_name
+        ]
+        logger.debug(f'app_table_names: {app_table_names}')
+
+        if spec_table_names is None:
```
pushyamig, line 74: Can we have a log info/debug message saying this case is entered? similarly for the else case. Info if it worth tracking in the prod logs

db/db_creator.py
```diff
@@ -18,18 +18,18 @@
...
 
 class DBCreator:
-
-    def __init__(
-        self,
-        db_params: Dict[str, str],
-        append_table_names: List[str] = []
-    ) -> None:
-
+    '''
```
pushyamig, line 23: Nice that you added  doc-strings for DBCreator, will be useful

db/db_creator.py
```diff
@@ -41,33 +41,60 @@ def __init__(
...
+        Gets table names using the SQLAlchemy Engine object.
+        '''
         logger.debug('get_table_names')
         return self.engine.table_names()
 
     def migrate(self) -> DBCreator:
+        '''
+        Updates database schema using yoyo-migrations and the migration files in the
+        migrations directory collocated with this file (db_creator.py).
+        '''
```
pushyamig, line 55: You are returning the `self`. I never used/knew this kind of technique. Looking up quickly tell me this part of the Fluent interface approach you using with DBCreator.  I think to add a sentence in the doc-string top near the `DBCreator` will make the approach you're taken with DbCreator

db/db_creator.py
ssciolla: Sure, I could do that.

db/db_creator.py
ssciolla: Okay, yes, I can add a clause or sentence to the top docstring stating that some methods can be used fluently.

db/db_creator.py
```diff
@@ -41,33 +41,60 @@ def __init__(
...
         conn = self.engine.connect()
         conn.execute('SET FOREIGN_KEY_CHECKS=0;')
-        for table_name in self.get_table_names():
-            if 'yoyo' not in table_name and table_name not in self.append_table_names:
-                logger.debug(f'Table Name: {table_name}')
-                conn.execute(f'DELETE FROM {table_name};')
-                logger.info(f'Dropped records in {table_name} in {self.db_name}')
+        for drop_table_name in drop_table_names:
+            logger.debug(f'Table Name: {drop_table_name}')
+            conn.execute(f'DELETE FROM {drop_table_name};')
```
pushyamig, line 90: Looking into the `sqlalchemy` library, the `execute` function will raise an exception ( in the unknown situation). This should be wrapped in try-catch. Similarly, in the other parts of code that are doing a DB transaction. This issue need not be solved here but can be taken as a separate issue

ssciolla: Yeah, was thinking about this a little. I'll do some testing and can resolve it here.

db/db_creator.py
ssciolla: I added some `INFO`-level logging as you requested, so I'm resolving this. See [this commit](https://github.com/tl-its-umich-edu/course-inventory/pull/129/commits/96079b7a94d06d5ba7efcd3cd0e3ca3f09de10d7).

db/db_creator.py
ssciolla: Okay, I updated that top docstring [here](https://github.com/tl-its-umich-edu/course-inventory/pull/129/commits/9ff82a7f427fe1fd1396c82e0234f8d9f61fbe57). Let me know if that helps.

db/db_creator.py
ssciolla: @pushyamig, okay, I was a little concerned about us trying to delete records from tables that don't exist when using `reset_database`, but because I'm relying on `self.engine.table_names`, we're only ever going to get actual tables passed to the `DELETE` query. When specific strings are passed, I'm doing validation.

Can you think of other cases where we would really need to catch an exception? `try/except` can be useful, but I prefer to not use unless necessary, since it can lead to some confusing output if the exceptions aren't specific enough.

db/db_creator.py
```diff
@@ -41,34 +42,62 @@ def __init__(
...
+            logger.info(f'Dropped records in {drop_table_name} in {self.db_name}')
         conn.execute('SET FOREIGN_KEY_CHECKS=1;')
```
pushyamig, line 95: @ssciolla if the DB traction failed for some here we want it to be handled. Otherwise, Foreign constraints will be effected. 

In general, I have seen that whenever we are writing data using pandas SQL Alchemy we are not surrounding with try-catch. we have to review all the cases with DB transaction and see is proper handling is implemented if not here but open an issue and deal with it later.

db/db_creator.py
pushyamig: https://github.com/tl-its-umich-edu/course-inventory/pull/129/files#r419535834

db/db_creator.py
ssciolla: Okay, so what do you mean by handled? Just to catch the exception and report it? How would that be different from just looking for an exception? I understand the possible concern of tables getting messed up, but that would happen regardless I think. MySQL doesn't support rollbacks of transactions. I'm just trying to understand what wrapping things in `try/except` accomplishes.

Also, do you see this is a a significant issue? If so, we should be addressing it before we release. I don't have a good sense of how big of a problem you see this as. We haven't talked about it before, even though we've worked on code together that uses `to_sql`.

ssciolla: I am seeing that MyLA does wrap things in `try/except` in `cron.py`, so I'm looking there for some explanation/clarity.

ssciolla: If one query failed, I don't think we'd want the others to continue, right? It's all interconnected.

pushyamig: Yeah.. that's the idea. If we log it as an error with try-catch then Jenkin sees that as error and reports it. At the minimum we might need to do is rerun the cron to fix it.

ssciolla: I still don't see how that would be different from just letting it error out; Jenkins would still log it as a failure.

I am seeing that there might be ways to support rollbacks with MySQL and Python, but I'm not sure yet if there's a way to do it with `pandas`.

Would you mind creating an issue for this? I think we can have this discussion there, and if we need to come up with an emergency fix, I can work on that.

pushyamig:
I was able to run this with out any issues

PR #133, opened 2020-05-01
zoom_courses.csv being written to wrong location

### LINKED ISSUE #132: zoom_courses.csv being written to wrong location
It looks like zoom_courses_meetings.csv is going to the right place but not zoom_courses. Looks like a small issue with the os.path.join.

jonespm:
Fixes #132


jonespm:
Not sure why I had a + instead of a , here!

jonespm:
They say every 100 lines of code will always have 1 bug that sneaks in! ;)

ssciolla:
No worries, I've been finding as I work on some parts of this project that stuff I wrote a couple weeks ago is just bad and needs to be changed (e.g. see issue #72 and PR #129). I guess it's an iterative process, and sometimes it's simpler to just merge and evaluate as you use the code more.

jonespm:
Yup for sure, [just get it in there](https://en.wikipedia.org/wiki/Perfect_is_the_enemy_of_good), reduce the conflicts for everyone else and get the small tweaks in after!

Issue #139, opened 2020-05-06
Remove unused dependencies from requirements.txt


ssciolla:
`requirements.txt` includes a couple of dependencies that were only used by the now removed Zoom code: `pyjwt`, `pyyaml`, and `dateparser`. These should also be removed.

jonespm:
I tested this, all of the imports that were removed are unused and I don't see anything that's imported that's not needed.

PR #136, opened 2020-05-04
Remove unneeded video conferencing code (Zoom)

jonespm:
Fixes #127

PR #140, opened 2020-05-06
Update requirements.txt (#139)

ssciolla:
This PR updates `requirements.txt` to remove dependencies no longer in use, bump versions for bugfix releases, pin `requests`, and alphabetize entries. More details are provided in the task list below. The PR aims to resolve issue #139.

- [x] Remove `pyyaml`, `pyjwt`, and `dateparser` (which were used in Zoom code, removed by PR #136).
- [x] Bump versions for `psycopg2-binary`, `pandas`, and `SQLAlchemy` to latest bugfix releases (I waited on `beautifulsoup4`, since that has a minor release that might require more testing; it's also still at `4.9.0`).
- [x] Add `requests` and pin to version `2.22.0`, which is compatible with `umich-api` (I was seeing a warning during the build; we'll need to update that in `api-utils-python` at some point).
- [x] Alphabetize entries and remove spacing.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,16 +1,13 @@
+beautifulsoup4==4.8.2
+canvasapi==0.15.0
+google-cloud-bigquery==1.24.0
 hjson==3.0.1
 jsonschema==3.2.0
 mysqlclient==1.4.6
-pandas==1.0.1
-psycopg2-binary==2.8.4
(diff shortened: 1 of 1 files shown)


ssciolla:
Thanks for the quick reviews; I'm merging now.

PR #143, opened 2020-05-07
implement `DataSourceStatus` (#142)

### LINKED ISSUE #142: replace data-source-timestamp dictionaries with objects
Methods that are called by `Job.run()` are expected to return a dictionary with the weakly-defined string keys "`data_source_name`" and "`data_updated_at`".  The types of those keys' values are also weakly-defined as `ValidDataSourceName` and `datetime`.

Add a class to replace this dictionary with well-defined properties and types.

lsloan:
* 🎟Issue: Resolves #142
* ⏰Review due: 15 May 2020, 13:00 

Admittedly, part of the motivation behind implementing this is that Sam said it would be tricky.  I didn't think it would be.

## Testing
The correct configuration values are available in the `env-lsloan_hjson.txt` file I uploaded to the `TL Security files` > `Reporting Project (course-inventory)` folder in Box.

## Review style
I **_humbly_** request that comments on this PR be nested under a single review.  It helps me if the comments are organized this way, especially if the comments request changes in the code.


lsloan:
@ssciolla and @jonespm, although this is PR is still a draft, I wanted to get your opinion on this.  I'm trying to decide whether I need to change all the parts of the code that had been using the data source dictionary or is it OK to leave it as it is, with dictionary backwards compatibility features?

The reason I've not gone for changing the other parts of the code already is that I thought there might be some objection to that.  I'd hate to do that work only to have to undo it.

lsloan:
As @ssciolla mentioned in item number 3 of his comment on my recent PR (https://github.com/tl-its-umich-edu/course-inventory/pull/116/#issuecomment-626010997), I will change the name of this new class from `DataSourceTimestamp` to `DataSourceStatus`.  It makes sense for the name of this class to match the name of the table in which this data is stored and it doesn't limit the content of the class to only timestamp.  It could contain more in the future.

However, for the next step, I'm uncertain which of these to do:

1. Keep the dictionary-compatibility code I've added to the class (`copy()` and `__getitem__()`, seen at https://github.com/tl-its-umich-edu/course-inventory/pull/143/files#diff-b7e82fd4a46659bb0ab7b4f46858cbdfR72 in this PR)
1. Remove the dictionary-compatibility code I've added to the class and update the code in `run_jobs.Job#create_metadata()` to use the objects properly.  This may also require me to change the code in the Canvas, Zoom, BlueJeans, etc. parts of the application.

One or the other needs to happen.

Sam's statement about this in item 3 of his comment is that I don't need to do either.  In order for me to test my code, I **_needed_** to do one of them, so I chose the first option, backwards compatibility.  My preference is to keep that, because to do the other option would involve my changing code in other parts of the application.  I don't mind doing that, but if the other authors want to maintain the code they wrote, I understand.

ssciolla:
@lsloan, I'm sorry, maybe my comment on the other PR wasn't clear. I said, 

`Also, I don't think you need to worry about backwards compatibility or changing other people's code. If you think this is a better way to do it, we should change it everywhere.`

I meant that I don't think backwards compatibility is the way to go AND that you should not be concerned or worried about changing other people's code, i.e. that you SHOULD change other people's code rather than supporting divergent processes (that could arguably make code harder to maintain). This is part of the code that everyone needs to use, and if you think your changes improve the process, then we should improve it everywhere. 

That's my view; maybe other's disagree.

lsloan:
Thanks for the clarification, @ssciolla.  That makes more sense than my interpretation of your earlier message.  I should have asked sooner.

I'll work on updating the various parts of the code to use `DataSourceStatus` and test them.

vocab.py
```diff
@@ -20,9 +28,69 @@ class ValidDataSourceName(Enum):
...
+        For backwards compatibility with code that expects to copy this object as a dictionary.
+
+        TODO: Update other code to not need this, then remove this method.
+
+        :return: Dict of object properties
+        '''
+        return {
+            'data_source_name': self.dataSource,
+            'data_updated_at': self.timestamp
+        }
```
ssciolla, line 85: @lsloan, I know you're still working on updating this, but I was thinking about it a little bit. You may still need something like this for it to be compatible with the use of `pd.DataFrame` in `run_jobs.py`. Even if not, it still be useful to have a `to_dict` or similar method for testing and logging purposes.

lsloan: Yes, you're right.  We're thinking along the same lines.  I ended up keeping `copy()`, with a slight change.

lsloan:
Apologies for the on/off/on review status.

@jonespm and @pushyamig, if you don't have time to review this PR, please communicate with the other reviewers or add a new reviewer.

@ssciolla, thanks for your feedback on this PR thus far.  I didn't request your review, but I welcome your comments.  As mentioned in the dev. meeting a few days ago, you're getting a lot of review requests.  Maybe you need a break from them.

pushyamig:
I am able to review this PR. I am taking me out from this

vocab.py
```diff
@@ -20,9 +28,92 @@ class ValidDataSourceName(Enum):
...
+
+        :param dataSourceName: `ValidDataSourceName` - Name of data source
+        :param dataUpdatedAt: `datetime` - Time at which data from the source was updated
+        '''
+        self.dataSourceName: ValidDataSourceName
+        self.dataUpdatedAt: datetime
+        self.setDataSourceName(dataSourceName)
+        self.setDataUpdatedAt(dataUpdatedAt)
+
+    def setDataSourceName(
```
jonespm, line 59: The preferred way to implement these in python is to use a [property decorator](https://docs.python.org/3/library/functions.html#property) and the getter/setter decorators for those.

lsloan:
> I am able to review this PR. I am taking me out from this

I think you mean "unable", right, @pushyamig? 😉

vocab.py
lsloan: Thanks for the suggestion.  I've not done it that way before, but I'll give it a try.

I've looked at the example you linked to.  I'm unsure whether in the end the methods are intended to be called as `instOfC.x = 5` and `instOfC.x` for set and get, respectively.  I understand that some may want to use properties that way, but I don't like it.  It also doesn't make a fluent interface possible.

jonespm: Yeah, that's basically how you'd do it.

[Python is not Java
](https://docs.quantifiedcode.com/python-anti-patterns/correctness/implementing_java-style_getters_and_setters.html)

Even in Java everything uses Lombok annotations now for these.

lsloan: I never thought I'd say it:  I'm starting to dislike Python.

lsloan: I'm going to take a hybrid approach.  I will follow the Pythonic techniques to define functions for getting and setting the properties.  However, I will keep the setter properties so the class can retain a fluent interface.

I believe Guido has said he's not in favor of fluent interfaces.  But I like them.  Even though he is BDFL, this is still a democracy.

vocab.py
```diff
@@ -20,9 +28,92 @@ class ValidDataSourceName(Enum):
...
+
+
+class DataSourceStatus:
+    '''
+    Each data source process must return an instance of this class, which tells the name of the
+    data source and the time at which data from it was last updated.
+    '''
+    def __init__(
+            self,
+            dataSourceName: ValidDataSourceName = NotImplemented,
```
jonespm, line 46: PEP-8 camelCase is [really only *allowed* to conform](https://david.goodger.org/projects/pycon/2007/idiomatic/handout.html#naming) to pre-existing conventions. Is there some reason you're preferring this here in instance variables but joined lower in other places?? 

Generally these conventions were for code that was converted over from another language.

lsloan: I believe I am consistent in using camelCase.  I don't know of any place I use `joined_lower` here.  If I did, please point it out.  I'll check for it again.

I use `JOINED_CAPS` when naming constants.

I've discussed this before with Sam.  He wasn't very enthusiastic about my preference, but he thought he could live with it if it's confined to specific module(s).

ssciolla: @lsloan, since this is a shared module that everyone will have to use and the prevailing style of this project is `snake_case`, I would prefer that you use that in this file, please.

jonespm: Yeah, I was going to basically say the same as that's the Python style. Unless this code was inspired from a module written in Java or if this was already the convention of the project it should be joined_lower. This is a shared module that others will contribute to.

**Naming**
* joined_lower for functions, methods, attributes
* joined_lower or ALL_CAPS for constants
* StudlyCaps for classes
* camelCase only to conform to pre-existing conventions
* Attributes: interface, _internal, __private

lsloan:
I've made some changes according to @jonespm's suggestions.  Using the Pythonic `property()` decorators for most of the class methods, but retaining the `set*()` methods to allow for chaining.

I found a couple of things to improve along the way, removing some unnecessary checking logic and cleanup according to various Pylint tools.

vocab.py
lsloan: I have a **_rant_** about PEP 8, but I'll not waste time with that here.

jonespm:
@lsloan I removed this from the project board because we have only been adding issues to the project board and #142 is already on there. Having the PR and issue both on there is confusing. You could use a label if you want to indicate some type of status on this.

lsloan:
> @lsloan I removed this from the project board because we have only been adding issues to the project board and #142 is already on there. Having the PR and issue both on there is confusing. You could use a label if you want to indicate some type of status on this.

Sorry.  I noticed it wasn't set just now, so I added it out of habit.  I'll remove it.

lsloan:
@jonespm and @ssciolla, I've updated my code from camelCase to snake_case.  Go ahead and review again.

All in all, if the biggest problem with my code is that I use camelCase for my variables, then I'm pretty happy with it. 😉

vocab.py
lsloan: I should say that I looked over these reviews this morning before 09:00.  I didn't look back until after @ssciolla mentioned in a meeting today that he specifically asked me to not use camelCase.  I didn't see his message when I committed some changes and prompted him and @jonespm to review again.  I wasn't purposely ignoring his request.  I guess I need to spend more time checking for GitHub comments in the GH UI and my email before committing changes and requesting reviews.

vocab.py
```diff
@@ -20,9 +28,102 @@ class ValidDataSourceName(Enum):
...
+        '''
+        if (data_updated_at is NotImplemented):
+            self._data_updated_at = pd.to_datetime(time.time(), unit='s', utc=True)
+        elif (data_updated_at.tzinfo is not pytz.UTC):
+            raise ValueError('_data_updated_at must have UTC time zone')
+        else:
+            self._data_updated_at = data_updated_at
+        return self
+
+    def copy(self) -> Dict:
```
ssciolla, line 119: Nit: could be more precise here with the type hint, e.g. `Dict[str, Union[ValidDataSourceName, pd.Timestamp, datetime]]`

vocab.py
ssciolla: No worries, sorry if I came on a little strong, totally reasonable that you missed the comment. I just figured I'd have a direct conversation with you and hear what your thinking was.

vocab.py
```diff
@@ -20,9 +28,102 @@ class ValidDataSourceName(Enum):
...
+
+class DataSourceStatus:
+    '''
+    Each data source process must return an instance of this class, which tells the name of the
+    data source and the time at which data from it was last updated.
+    '''
+
+    def __init__(
+            self,
+            data_source_name: ValidDataSourceName = NotImplemented,
```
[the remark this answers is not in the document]
lsloan: I recently picked up this usage of `NotImplemented` from another module.  I think it was the Kaltura API.  

It appears to be used there because `None` could be a valid value of the arguments.  `NotImplemented` is used to indicate a value wasn't given for an argument.

However, the more I think about this, it's obvious that `None` is not a valid value for this code, so there's no reason not to use that.  Also, probably `False` would be a better indicator than `NotImplemented` in a lot of other cases.

For the sake of harmony, I'll change the default values here to `None`.  Then I could set the type hint to `Optional[ValidDataSourceName]`, although I really, really dislike the name `Optional` for this purpose, but I also dislike the wordiness of a `Union` that includes `None`.

In these cases, since `None` is just a placeholder to indicate a value wasn't given, I feel like the best hint is the **_desired_** type.

vocab.py
```diff
@@ -20,9 +28,102 @@ class ValidDataSourceName(Enum):
...
+            data_updated_at: datetime = NotImplemented
```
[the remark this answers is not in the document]
lsloan: I've changed this section of code to not use Pandas.

As you point out, `pd.Timestamp` objects can often be used like `datetime` ones, thanks to Python duck-typing.  So that other type can be passed as an argument and it will work.

vocab.py
lsloan: Removed Pandas dependency.

lsloan:
I answered a lot of @ssciolla's questions in the individual comments, but there are a couple things from the overall approval comment that I skipped over.

> I could have made a couple more about missing `--> None` return hints.

I noticed this comment on my first read and I meant to do something about it, but I forgot about it as I began addressing the other issues.  Yes, I should have added `-> None` hints to `@data_source_name.setter` and `@data_updated_at.setter`.

BTW, have you noticed that when using `property()` decorators, and the methods all need to be named the same, it's no longer convenient to refer to them in conversation by their names?  If I had said I need to make some change to `data_source_name()`, to which am I referring?  There's a getter and a setter.  I ended up referring to these by their decorators.

> I find all the different setter methods to be unnecessarily convoluted, but I believe some of that feeling comes from being trained originally in Python. Thank you @lsloan for acquiescing to my request to use `snake_case`. I will try to be better about actually using "Request changes".

I agree, but from a different perspective.  I don't like that when I used the `property()` decorators, I needed to write stub setters because I wanted to support method chaining.  Maybe I should have written this using `property()` as a function rather than decorators.  My code would probably be less redundant.  I do really like the look of the decorators, though.  So I have mixed feelings.

| Note |
| :--- |
| I'd referred to method chaining before as "fluent interface" (FI).  I've since remembered that FI should also involve a domain-specific language (DSL).  In this case, and possibly many others, there's no need for a DSL, but chaining methods is convenient. |

Issue #145, opened 2020-05-08
Possibility of duplicate primary key after dropping duplicates in AsyncEnrollGatherer


ssciolla:
We encountered an error today when running the test job where the `user` record insertion failed because of a duplicate primary key. Based on our investigation, it seems that a `user` record might have changed during the `AsyncEnrollGatherer` process, resulting in two `user` records with two non-duplicate records with the same `canvas_id`.

We need to fix this so that this outlier case is handled. We can do this by changing the code in `generate_output` to use the `subset` parameter of `pd.DataFrame.drop_duplicates()` so that we're only considering the primary key. We should also use the `keep` parameter, specifying `last` so that whatever the latest record is is maintained.

It probably also makes sense to consider these changes for `enrollment` and `section`.

https://github.com/tl-its-umich-edu/course-inventory/blob/2e27243e018180e38d049747500d6c93129ad91c/course_inventory/async_enroll_gatherer.py#L177-L186

PR #146, opened 2020-05-08
Fix duplicate dropping in AsyncEnrollGatherer (#145)

ssciolla:
This PR makes small modifications to the usages of [`pd.DataFrame.drop_duplicates`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.drop_duplicates.html) in `AsyncEnrollGatherer.generate_output()` to avoid the possibility of duplicate primary keys (which can occur if a name or other field changed during the process, resulting in two slightly different records before duplicates are dropped). This possibility was avoided with the use of `subset=['canvas_id']` as a parameter to `drop_duplicates`, which will mean that only `canvas_id` (the PK) is considered when identifying duplicates. In addition, to ensure the latest changes are kept, the parameter `keep='last'` was also added. These changes were made to `enrollment` and `section` as well as `user` out of an abundance of caution. The PR aims to resolve issue #145.

### DIFF
--- a/course_inventory/async_enroll_gatherer.py
+++ b/course_inventory/async_enroll_gatherer.py
@@ -177,9 +177,9 @@ class AsyncEnrollGatherer:
         # Seems like we shouldn't have to drop duplicates for enrollments, but once one
         # duplicate broke the process
         enrollment_df = pd.DataFrame(enrollment_records)
-        enrollment_count = len(enrollment_df)
+        orig_enrollment_count = len(enrollment_df)
         enrollment_df = enrollment_df.drop_duplicates(subset=['canvas_id'], keep='last')
-        logger.info(f'{len(enrollment_df) - enrollment_count} enrollment records were dropped')
+        logger.info(f'{orig_enrollment_count - len(enrollment_df)} enrollment records were dropped')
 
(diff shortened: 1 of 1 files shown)


ssciolla:
FYI, I ran the `COURSE_INVENTORY` job for the Spring and Summer semesters without this change and with this change. Record numbers for the affected tables were identical. This is quite difficult to test, but a lack of downstream data changes is somewhat reassuring.

ssciolla:
Okay, I did some testing and more or less everything worked the way I expected. You can see the testing code in [this commit](https://github.com/ssciolla/course-inventory/commit/fe0ac6f5d56c14020af73bf4d7b9d161b788d2d0), but I've also linked it below: 

https://github.com/ssciolla/course-inventory/blob/fe0ac6f5d56c14020af73bf4d7b9d161b788d2d0/course_inventory/async_enroll_gatherer.py#L177-L209

ssciolla:
Also, I discovered in the line where I report the number of enrollment records dropped, I had the counts out of order, so it was showing a negative number; I switched them.

ssciolla:
Okay, I'll merge this. Thanks for the reviews. The testing helped me feel more confident about this.

Issue #138, opened 2020-05-05
Fetch account data, store in new account table


ssciolla:
From the Trello card created by @jennlove-um:

"The unit hierarchy in reports should be based on the UM department hierarchy used in other reporting. To do this, we need to map Canvas account_id for courses to a department ID.

The Canvas account_dim has the sis_source_id which maps to the UM deptid.  

We need to add this account data to our Canvas data for the dashboard.  We should pull the canvas_id, name, and sis_source_id from the account_dim."

The account_dim refers to UDW, but I think it would be simple and better long term to get this directly from the Canvas API. The returned `account` object and available endpoints are described [here](https://canvas.instructure.com/doc/api/accounts.html). It looks like there are probably only 100-200 different account IDs in this dataset, so either the "List accounts" or "Get single account" endpoint should work.

zqian:
There are 147 account entries in the db now. This is not a complete set of all Canvas subaccount. I guess the reason for that is only the account with course/sections are retrieved?

ssciolla:
@zqian, yeah, I designed it to only care about accounts related to a course. I could change it to pull all accounts, if you want. Let me know!

jennlove-um:
@zqian My inclination is to keep pulling only the accounts that have courses in them.  There are hundreds of accounts that are unlikely to ever have courses associated with an academic term, and that feels like adding noise to the data for little to no purpose.  For example, there are over 700 accounts in the Non-Academic Program Courses account alone.  Unless we get a specific request and business need to include all accounts, I think it makes sense to include only accounts that have courses in one of the academic terms included in the data and the visualizations.

zqian:
@ssciolla @jennlove-um I got the idea of reporting only the data needed. Is it possible that going into future term, there will be a new course associated with a new account, not reported previously in the Account table? Will the report detect the new account and add it to the account table?

ssciolla:
Yes, with the current set up, all accounts associated with courses will be available. This is transparent in the code; unique account IDs found in the course data are fed into the account data fetching function, which specifically requests data for each account ID.

https://github.com/tl-its-umich-edu/course-inventory/blob/27383d09ba47f7fa6f0ad608bbe3b46141bae077/course_inventory/inventory.py#L278-L279

How exactly this works may be modified if or when we move towards updating (rather than dropping records), but we can ensure new accounts are handled appropriately.

zqian:
Sounds good. I will move this the Done column in the project.

PR #151, opened 2020-05-12
Fetch account data from Canvas API, create and populate account table (#138)

ssciolla:
This PR expands `course_inventory/inventory.py` to model, gather, and store Canvas account data, namely, `account.canvas_id`, `account.sis_id`, and `account.name`. It introduces a new function gathering account data that accepts a list of account IDs (pulled from associated course data) and returns a `pandas.DataFrame`, which is then written to CSV (optionally) and the database. The existing `course.account_id` field was also made a foreign key, connecting to `account.canvas_id`. The PR aims to resolve issue #138.


pushyamig:
I had local DB with the latest migration changes from the master and after the I pulled in this PR and ran the process again. I see below error
```
course_inventory | INFO:__main__:MySQL caught up
course_inventory | INFO:__main__:Applying any new migrations
course_inventory | INFO:yoyo.migrations:Applying 0016.add_account_table
course_inventory | INFO:yoyo.migrations: - applying step 0
course_inventory | INFO:yoyo.migrations: - applying step 1
course_inventory | INFO:yoyo.migrations: - rolling back step 0
course_inventory | Traceback (most recent call last):
course_inventory |   File "/usr/local/lib/python3.7/site-packages/yoyo/migrations.py", line 313, in apply
course_inventory |     getattr(self.step, direction)(backend, force)
course_inventory |   File "/usr/local/lib/python3.7/site-packages/yoyo/migrations.py", line 413, in apply
course_inventory |     self._execute(cursor, self._apply)
course_inventory |   File "/usr/local/lib/python3.7/site-packages/yoyo/migrations.py", line 379, in _execute
course_inventory |     cursor.execute(stmt)
course_inventory |   File "/usr/local/lib/python3.7/site-packages/MySQLdb/cursors.py", line 209, in execute
course_inventory |     res = self._query(query)
course_inventory |   File "/usr/local/lib/python3.7/site-packages/MySQLdb/cursors.py", line 315, in _query
course_inventory |     db.query(q)
course_inventory |   File "/usr/local/lib/python3.7/site-packages/MySQLdb/connections.py", line 239, in query
course_inventory |     _mysql.connection.query(self, query)
course_inventory | MySQLdb._exceptions.IntegrityError: (1452, 'Cannot add or update a child row: a foreign key constraint fails (`course_inventory_local`.`#sql-1_3`, CONSTRAINT `fk_account_id` FOREIGN KEY (`account_id`) REFERENCES `account` (`canvas_id`) ON DELETE CASCADE ON UPDATE CASCADE)')
```
and with additional stack trace as 
```
Traceback (most recent call last):
course_inventory |   File "run_jobs.py", line 136, in <module>
course_inventory |     db_creator_obj.migrate()
course_inventory |   File "/app/db/db_creator.py", line 61, in migrate
course_inventory |     backend.apply_migrations(backend.to_apply(migrations))
course_inventory |   File "/usr/local/lib/python3.7/site-packages/yoyo/backends.py", line 435, in apply_migrations
course_inventory |     self.apply_migrations_only(migrations, force=force)
 File "/usr/local/lib/python3.7/site-packages/MySQLdb/connections.py", line 239, in query
course_inventory |     _mysql.connection.query(self, query)
course_inventory | MySQLdb._exceptions.OperationalError: (1305, 'SAVEPOINT sp_2 does not exist')
```

course_inventory/inventory.py
```diff
@@ -182,6 +182,31 @@ def gather_course_data_from_api(account_id: int, term_ids: Sequence[int]) -> pd.
     return course_df
 
 
+def gather_account_data_from_api(account_ids: Sequence[int]) -> pd.DataFrame:
+    logger.info('** gather_account_data_from_api')
+    url_ending_with_scope = f'{API_SCOPE_PREFIX}/accounts/'
+
+    account_dicts = []
+    for account_id in account_ids:
+        logger.info(f'Account number {account_id}')
```
pushyamig, line 191: The message can say something like 'Fetching data for account {account_id}`

ssciolla:
> I had local DB with the latest migration changes from the master and after the I pulled in this PR and ran the process again.

@pushyamig, can you wipe out your database (delete `.data`) and try this again? I think having data in the tables will cause this to fail because of the new relationship to `account`.

course_inventory/inventory.py
ssciolla: I wasn't sure if this was necessary given that I'm logging the function name which describes this. Could I just add one logging message about this `for` loop that says "Fetching account data"?

pushyamig:
> @pushyamig, can you wipe out your database (delete .data) and try this again? I think having data in the tables will cause this to fail because of the new relationship to account.

Then we need to do the same process in test and prod process once this gets merged. 
isn't migrations should handle that?

ssciolla:
> > @pushyamig, can you wipe out your database (delete .data) and try this again? I think having data in the tables will cause this to fail because of the new relationship to account.
> 
> Then we need to do the same process in test and prod process once this gets merged.
> isn't migrations should handle that?

Yes, we need to do that when we roll out changes to the OpenShift processes. Migrations don't know what the data is in the `account` table (or the `term` table; we did this for the release, so the process is documented). We are not currently dropping records before doing migrations, because that would mean the database would be empty for an hour or more while the job finishes.

pushyamig:
ok, after deleting the `.data` I can see data is populated correctly and the job ran fine
Interesting that the root account in canvas doesn't have a sis id int he UMdept table.

pushyamig:
I think in Django it applies migrations first and then delete and load happen. I don't think we never encountered an issue for the Myla process or we stepped into those cases yet. Maybe we don't have strict Foreign key constraints for some of the tables.

course_inventory/inventory.py
```diff
@@ -182,6 +182,31 @@ def gather_course_data_from_api(account_id: int, term_ids: Sequence[int]) -> pd.
...
+        account_url_ending = url_ending_with_scope + str(account_id)
+        response = make_request_using_api_utils(account_url_ending)
+        account_data = json.loads(response.text)
```
pushyamig, line 194: The JSON load can raise an exception. should handle it

ssciolla: This is already being handled in `make_request_using_api_utils`: https://github.com/tl-its-umich-edu/course-inventory/blob/aa9bfa7e1307ed564df16a8eb5b3b487fbc8ffae/course_inventory/inventory.py#L49-L74

ssciolla: @pushyamig, does this not resolve your concern? Or maybe you just didn't see the comment?

lsloan:
If this isn't already merged by then, I'll have time to review this on Friday morning.

course_inventory/inventory.py
ssciolla: Do we want it to continue if all three attempts fail? I'm logging an error and return a response for consistency, but I was figuring that the code would break at that point, which is probably what we want. If we want to use `sys.exit(1)` inside the function, I'm okay with that. We haven't encountered this case yet in my memory.

ssciolla: Depending on your answer, I think this might be best addressed in a separate issue. I'm not modifying this function at all in this PR.

course_inventory/inventory.py
ssciolla: I think I've resolved your concerns here with [this commit](https://github.com/tl-its-umich-edu/course-inventory/pull/151/commits/25f0043f5d438f9c33aba909bd3b8a9abcad1788).

course_inventory/inventory.py
```diff
@@ -182,6 +182,31 @@ def gather_course_data_from_api(account_id: int, term_ids: Sequence[int]) -> pd.
     return course_df
 
 
+def gather_account_data_from_api(account_ids: Sequence[int]) -> pd.DataFrame:
+    logger.info('** gather_account_data_from_api')
+    url_ending_with_scope = f'{API_SCOPE_PREFIX}/accounts/'
```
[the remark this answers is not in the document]
ssciolla: No other methods are using this endpoint. Are you saying that you want me to put it in all caps?

course_inventory/inventory.py
ssciolla: I think I had this in my original implementation, but when using the paging functionality of `umich_api`, you need to pass a response object to `API_UTIL.get_next_page`. See this code from `gather_course_data_from_api`: https://github.com/tl-its-umich-edu/course-inventory/blob/d4984bc91c0253955ec3b8b146e8882d203f5229/course_inventory/inventory.py#L149-L166

course_inventory/inventory.py
```diff
@@ -182,6 +182,31 @@ def gather_course_data_from_api(account_id: int, term_ids: Sequence[int]) -> pd.
     return course_df
 
 
+def gather_account_data_from_api(account_ids: Sequence[int]) -> pd.DataFrame:
+    logger.info('** gather_account_data_from_api')
```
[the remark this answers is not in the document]
ssciolla: It would be nice to see which method is producing the log message, but this also seems like it could lead to a lot of repetitive code if you're doing it in every method. Also, I agree, it would be nice to get the name of the method dynamically rather than hardcode it as a string. It sounds to me like this is a separate issue, and not one I want to address in this PR.

course_inventory/inventory.py
```diff
@@ -182,6 +182,31 @@ def gather_course_data_from_api(account_id: int, term_ids: Sequence[int]) -> pd.
     return course_df
 
 
+def gather_account_data_from_api(account_ids: Sequence[int]) -> pd.DataFrame:
```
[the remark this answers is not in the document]
ssciolla: Can you help me understand why this is necessary? I am using similar patterns throughout this file, so if I'm doing something wrong or unsafe, I'd like to know why. Account IDs when they are found on their own, in my view, should be an integer, since they need to be integers to be valid. Yes, I'm only using this as a string here, but relaxing the type hint doesn't make sense to me, since only valid IDs (integers) will help build a valid URL. I'm not seeing why every function/method needs to thoroughly validate its inputs; I don't think you're doing this in `extract.py`.

course_inventory/inventory.py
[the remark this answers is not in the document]
lsloan: Yeah, I would.

course_inventory/inventory.py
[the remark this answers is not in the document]
lsloan: It's a single line of repeated code.  It'd be nice to have a decorator for this purpose.

course_inventory/inventory.py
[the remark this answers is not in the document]
lsloan: As we discussed in Slack, this probably isn't a big deal.  This is an internal function, not part of a public API.  Using `str()` on the values in the function probably handles 99% of the problems with receiving the wrong data type.  The only place this function is called uses code that would seem to always return a sequence.

course_inventory/inventory.py
lsloan: OK.  Makes sense.

course_inventory/inventory.py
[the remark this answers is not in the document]
ssciolla: Even though this is only used in the local function scope? I tend to use ALL_CAPS for global variables and basically nothing else.

course_inventory/inventory.py
ssciolla: @pushyamig, I've opened issue #161 so we can continue discussing what you raised here.

course_inventory/inventory.py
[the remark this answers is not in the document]
ssciolla: I'm going to defer making this change, since it would probably require changes elsewhere in this file for consistency. Also, I think I'd like to have this discussion with you and other developers, since I've never done variable naming this way outside of the module level. Relevant link: https://www.python.org/dev/peps/pep-0008/#constants

ssciolla:
Thanks for the reviews and approvals! I'm merging now.

Issue #152, opened 2020-05-13
Remove user table to minimize PII collection


ssciolla:
After discussion and back-and-forth with various stakeholders, we've determined that the `user` table is not necessary for current reporting purposes, and that it should be removed to minimize Personally Identifiable Information (PII) collection. For now, the `user_id` in the `enrollment` table should be maintained to assist with generating student enrollment counts. Changing this will require a migration dropping the table and several modifications to `AsyncEnrollGatherer`.

zqian:
verified user table is no longer in database, and cron job works fine after the change.

PR #153, opened 2020-05-13
Remove user data collection and table (#152)

ssciolla:
This PR makes changes to the `COURSE_INVENTORY` job in order to remove and otherwise minimize collection of user-related data (PII). This is accomplished by modifying the existing `AsyncEnrollGatherer` class and its associated GraphQL query and adding a migration that drops the `user` table and an associated foreign key. More details are provided below in the task list. The PR aims to resolve issue #152.

- [x] Add migration removing the foreign-key constraint from the `enrollment` table and dropping the `user` table.
- [x] Remove all code dealing with `user` from the `AsyncEnrollGatherer` class and the `unnest_enrollment` function.
- [x] Remove all `user` fields except `_id` from the GraphQL query used by `AsyncEnrollGatherer`.
- [x] Add type hints to instance variable setting in`AsyncEnrollGatherer`.
- [x] Update usage of `AsyncEnrollGatherer`.
- [x] Remove `pull_sis_user_data_from_udw` and subsequent `pd.merge` usage.
- [x] Remove output code for `user`.
- [x] Make naming of `course_section` CSV consistent with table name.

### DIFF
--- a/db/migrations/0017.remove_user_table.py
+++ b/db/migrations/0017.remove_user_table.py
@@ -12,5 +12,5 @@ steps = [
     '''),
     step('''
         DROP TABLE user;
-    '''),
+    ''')
 ]
--- a/db/migrations/0017.remove_user_table.py
+++ b/db/migrations/0017.remove_user_table.py
@@ -1,5 +1,5 @@
 #
-# file: migrations/0015.remove_user_table.py
+# file: migrations/0017.remove_user_table.py
 #
 from yoyo import step
 
@@ -8,7 +8,7 @@ __depends__ = {'0006.modify_user'}
 steps = [
     step('''
         ALTER TABLE enrollment
-        DROP FOREIGN KEY fk_user_id;
+            DROP FOREIGN KEY fk_user_id;
     '''),
     step('''
         DROP TABLE user;
--- a/course_inventory/async_enroll_gatherer.py
+++ b/course_inventory/async_enroll_gatherer.py
@@ -1,6 +1,6 @@
 # standard libraries
 import copy, json, logging
-from typing import Dict, Sequence, Tuple
+from typing import Any, Dict, Sequence, Tuple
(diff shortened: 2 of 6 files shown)


db/migrations/0017.remove_user_table.py
```diff
@@ -0,0 +1,16 @@
+#
+# file: migrations/0017.remove_user_table.py
```
ssciolla, line 2: This is `17` in anticipation of PR #151 being merged.

jonespm: I feel like it would have been nice if we could have separated out these migrations that don't have any dependencies to their own directories.

Like we could have had 
mivideo/migrations
online_meetings/migrations
course_inventory/migrations

Then the migrations with no prior dependencies wouldn't wait on others. I'm working on one right now that I'm just calling 18_ to stay in line. Maybe something to fix if we ever have time in the future. ;)

ssciolla: Mmm, I see your point. I think I nixed this idea in the past because I thought it would unnecessarily complicate the codebase. Having a couple extra commits to change numbers when you're working at the same time as someone else felt like a small price to pay for less directories and keeping `db_creator.py` simpler. There probably is a way to do it though and maybe it wouldn't be too disruptive.

jonespm: Yeah, I know Django ORM does it by having every class extend migrations.Migration.

But we're doing it manually in the db_creator -> migrate -> read_migrations. 

We'd either have to hardcode all the paths (probably not too hard) to all the migrations or have some type of similar discovery process. 

Still I don't think it's a high priority.

ssciolla:
I pushed the migration number back to `0016` so I can merge this. I'll bump up the number in the other PR, #151. Thanks everyone for the reviews.

Issue #149, opened 2020-05-12
Refactor CSV/DB writing with logging in course_inventory


ssciolla:
The code that writes data to CSV or the DB and logs before and after is quite repetitive and can likely be made into function(s) or class(es) to reduce duplication. Note that this is pretty closely tied to whatever method we chose to use for writing to the DB (`pandas`, `SQLAlchemy`, etc.), so some consideration should be given to whether the current method is what we will use going forward.

PR #155, opened 2020-05-13
Write Canvas Meetings LTI to Database

### LINKED ISSUE #1: Fill out README
Add description and installation and usage instructions to README

### LINKED ISSUE #100: Fields need to be dropped from data

### LINKED ISSUE #137: Write Canvas Meetings LTI to Database
We need to change the to_csv in the canvas_zoom_meetings to write the to the database

This will involve

- [x] Creating the database schema that this will use
- [x] Writing the dataframes to the database

It may also be nice if it can update rather than delete and insert. This will require some additional research. This somewhat depends on #49 and would be improved by #114

jonespm:
- [x] Add conversion script
- [x] Change code to write to the database in the new schema format Fixes #137
- [x] Drop columns that shouldn't be in the datamodel Fixes #100 
- (Optional) Link lti_placement up to course table (This will be fixed in another issue) #162


jonespm:
@ssciolla This is still draft and you can wait to review or assign someone else.

Just adding tables for this process and the schema can change. My idea is to try to make it a little generic so we can run it on other LTI placements, but it's likely every placement will need a specific table to capture its details. 

We'll have 
*lti_placement_types* which is a just a dictionary to define the placements and the tables that hold the details

*lti_placement* is the actual LTI placement details. Ideally it will also link back to the course table instead of having accound_id, course_id and course_name but I think this *could* come later. This has a 1:1 with lti_placement_types. The values in this table are pulled from the Canvas API. 

*lti_zoom_placement* is a new table with zoom specific placement information with a *:1 back to lti_placement. The values here are pulled from the Zoom LTI API.


Going to try to implement this on the canvas_zoom_meetings code tomorrow. 

I numbered this 18_ as I think it will be behind some other commits.

Thanks!

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
...
+        INSERT IGNORE INTO lti_placement_types
+        (id, description, table_name) VALUES
+        (0, 'Zoom', 'lti_zoom_placement')
+    '''),
+    step('''
+        CREATE TABLE IF NOT EXISTS lti_placement (
+            id BIGINT NOT NULL AUTO_INCREMENT,
+            account_id INTEGER NOT NULL,
+            course_id INTEGER NOT NULL,
+            course_name VARCHAR(200),
```
ssciolla, line 26: Yeah, I think we could toss `account_id` and `course_name` if it's a Canvas course ID that can be linked back to the Canvas dataset. Like with the MiVideo stuff, we can't enforce a relationship because records are dropped, but you can still join.

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
...
+            placement_type INTEGER NOT NULL,
+            PRIMARY KEY (id)
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
```
ssciolla, line 29: `FOREIGN KEY` using `placement_type` to `lti_placement_type(s)`?

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
...
+    '''),
+    step('''
+        CREATE TABLE IF NOT EXISTS lti_zoom_placement (
```
ssciolla, line 32: We might want to think more about this table name. I think really what we're describing with each record is an `lti_zoom_meeting`?

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
...
+            id BIGINT NOT NULL AUTO_INCREMENT,
+            placement_id INTEGER NOT NULL,
+            meeting_id VARCHAR(100) NOT NULL,
+            host_id VARCHAR(100) NOT NULL,
+            topic VARCHAR(255) NOT NULL,
+            start_time DATETIME NOT NULL,
+            status INTEGER,
+            PRIMARY KEY (id)
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
```
ssciolla, line 41: `FOREIGN KEY` using `placement_id` to `lti_placement`?

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
...
+    step('''
+        CREATE TABLE IF NOT EXISTS lti_placement_types (
+            id INTEGER NOT NULL AUTO_INCREMENT,
+            description VARCHAR(100) NOT NULL,
+            table_name VARCHAR(64) NOT NULL,
+            PRIMARY KEY (id)
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
+    '''),
+    step('''
+        INSERT IGNORE INTO lti_placement_types
```
zqian, line 17: Is "ignore" really needed here?

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
...
+        (id, description, table_name) VALUES
+        (0, 'Zoom', 'lti_zoom_placement')
+    '''),
+    step('''
+        CREATE TABLE IF NOT EXISTS lti_placement (
+            id BIGINT NOT NULL AUTO_INCREMENT,
+            account_id INTEGER NOT NULL,
+            course_id INTEGER NOT NULL,
```
zqian, line 25: course_id shall be a FK to course table

db/migrations/0018_add_lti_usage.py
ssciolla: `placement_type_id` might be clearer?

db/migrations/0018_add_lti_usage.py
zqian: Is account_id really needed here?

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
...
+            course_name VARCHAR(200),
+            placement_type INTEGER NOT NULL,
+            PRIMARY KEY (id)
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
+    '''),
+    step('''
+        CREATE TABLE IF NOT EXISTS lti_zoom_placement (
+            id BIGINT NOT NULL AUTO_INCREMENT,
+            placement_id INTEGER NOT NULL,
```
zqian, line 34: placement_id is the FK to lti_placement table, right?

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
...
+'''
+
+from yoyo import step
+
+steps = [
+    step('''
+        CREATE TABLE IF NOT EXISTS lti_placement_types (
+            id INTEGER NOT NULL AUTO_INCREMENT,
+            description VARCHAR(100) NOT NULL,
+            table_name VARCHAR(64) NOT NULL,
```
ssciolla, line 12: I'm not sure this `table_name` is needed. Need to think more about this.

db/migrations/0018_add_lti_usage.py
jonespm: This is some of the disconnect between the scripts and having different developers work on different parts with no overall framework.

MiVideo also has mivideo_media_courses that doesn't link back to the courses table

The course table is only populated by the inventory script. We'd have to refactor to allow any of these scripts to add courses to the courses table. The problem is the inventory script when it runs drops all the courses so that would be problematic. This sounds like a much bigger issue.

account_id is here because it comes from the api. I could remove it. All of these could be removed if we could get a FK link to courses.

db/migrations/0018_add_lti_usage.py
jonespm: Yeah right, I didn't set up FK's

db/migrations/0018_add_lti_usage.py
jonespm: I only left it in there because we have a pattern of `CREATE TABLE IF NOT EXISTS`. We wouldn't want to insert twice.

The migrations would prevent this from running so we *really* don't need the INSERT IGNORE or IF NOT EXISTS.

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
+'''
+Migration for MiVideo media started hourly table
```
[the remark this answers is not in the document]
jonespm: Yeah, I copied another migration as a template to start with and left this in.

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
...
+'''
+
+from yoyo import step
+
+steps = [
+    step('''
+        CREATE TABLE IF NOT EXISTS lti_placement_types (
```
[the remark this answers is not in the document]
jonespm: Yeah, I noticed this after I committed was going to read about best practices, I can change this.

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,43 @@
...
+            id INTEGER NOT NULL AUTO_INCREMENT,
+            description VARCHAR(100) NOT NULL,
+            table_name VARCHAR(64) NOT NULL,
+            PRIMARY KEY (id)
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
+    '''),
+    step('''
+        INSERT IGNORE INTO lti_placement_types
+        (id, description, table_name) VALUES
+        (0, 'Zoom', 'lti_zoom_placement')
```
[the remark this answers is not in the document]
jonespm: Yeah, I switched this because I wanted the explicit 0 for this row to be able to reference this in the code later. I could remove the auto increment or just trust that it the order will be predictable.

I only have the IGNORE it in there because we have a pattern of `CREATE TABLE IF NOT EXISTS`. We wouldn't want to insert twice.

The migrations would prevent this from running so we *really* don't need the INSERT IGNORE or IF NOT EXISTS.

db/migrations/0018_add_lti_usage.py
jonespm: Yes, if we had a way to populate this course table easier I'd have done this but I figured it didn't need to be part of this change. The only place I saw it populated was in inventory.py on or near line 365.

A method to actually populate to the course tables doesn't exist as part of this code yet independent of this and I didn't see where/if MiVideo was doing this differently. I would actually want to call some utility method that would create the course if it doesn't exist, then be able to link to that.

db/migrations/0018_add_lti_usage.py
jonespm: Okay will change this

db/migrations/0018_add_lti_usage.py
jonespm: I have this in here because we either have the table names in the code or we have this in the database.

This entire table could be in the code too instead of the database.

db/migrations/0018_add_lti_usage.py
ssciolla: I understand your thinking here, but I don't know if we want multiple processes to be responsible for creating course records. We did talk briefly in the past about the possibility of making your script dependent on `course_inventory`, i.e. fetch the course data from the database rather than running a separate process; then you'd know that the `canvas_id` really was in `course`, even if we couldn't enforce it.

db/migrations/0018_add_lti_usage.py
jonespm: I'm somewhat leaning on deleting this table and just doing the mapping in the code and when/if we add placement(s) other than Zoom doing it then.

db/migrations/0018_add_lti_usage.py
jonespm: Yeah, that's probably a more correct name. 

Back when we were running the Zoom report it was going to be a direct link to the meetings pulled from the Zoom API. Now, well it's just a stand alone table since that information is likely in a different database.

jonespm: I can change this name though

lsloan:
If this isn't already merged by then, I'll have time to review this on Friday morning.

db/migrations/0018_add_lti_usage.py
ssciolla: Yeah, so far, I've taken the somewhat lazy approach of not defining controlled vocabularies in the database (for enrollment types, for example). I kind of figured that this was an intermediary data source, and that the DWing team would follow best practices to create an analytical dataset.

Regarding the `table_name`, yeah, it might make more sense to include this in the codebase as an `enum`. I'm just thinking that the `table_name` here is describing a related entity (a Zoom meeting, a ProctorU test, etc.), and thus it feels a little out of place in this `lti_placement_type`.

db/migrations/0018_add_lti_usage.py
jonespm: Sure but that's the way it is now. It's also possible that this may pull down different courses than the course_inventory since course_inventory has some other criteria and this will eventually just pull from the report.

We could refactor this whole thing and have

A top level job that processes though courses
- Get courses in a term that match the criteria
  - Runs the course inventory
- Get courses from the LTI placement report
  - Runs the LTI placements

Or we could just have some method that creates courses for all of these scripts. I don't know how MiVideo process would fit into this courses job?

jonespm: I think we should just create a new issue to deal with the course FK problem as it's going to be bigger than getting this data into the database. 

I agree that multiple processes shouldn't be creating this but I don't know how a single process can easily accommodate for everything that may need to use this without having a pre-process separate from either of these before to update the courses (without dropping it).

ssciolla: Sounds fine to me. Do you want to create the issue or do you want me to?

jonespm:
Thanks Lance, it's not going to be ready till at least Monday.

db/migrations/0018_add_lti_usage.py
jonespm: I'm creating a new issue to deal with this

db/migrations/0018_add_lti_usage.py
jonespm: I'll create this

db/migrations/0018_add_lti_usage.py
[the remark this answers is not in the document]
jonespm: I'm just going to delete this table and do the enum

db/migrations/0018_add_lti_usage.py
jonespm: I created an issue #162 to resolve this

db/migrations/0018_add_lti_usage.py
jonespm: I'm just deleting this table and doing it in the code

jonespm:
Got a caffeinated power session so finished this today! ;) 

If you want to test this you have to generate a Canvas Api Token under an admin account. I'll generate one under a generic account we've been working on this week and put that in box.

db/migrations/0018_add_lti_usage.py
```diff
@@ -0,0 +1,30 @@
...
+        CREATE TABLE IF NOT EXISTS lti_zoom_meeting (
+            id BIGINT NOT NULL,
+            course_id INTEGER NOT NULL,
+            meeting_id VARCHAR(100) NOT NULL,
+            host_id VARCHAR(100) NOT NULL,
+            topic VARCHAR(255) NOT NULL,
+            start_time DATETIME NOT NULL,
+            status INTEGER,
+            PRIMARY KEY (id),
+            FOREIGN KEY (course_id) REFERENCES lti_placement(id) ON DELETE CASCADE ON UPDATE CASCADE
```
ssciolla, line 27: Is using `course_id` as the foreign key going to work if or when we have multiple LTI types in `lti_placement`? Seems like this would create some inaccurate joins. Is it possible to use the compound key?

online_meetings/canvas_zoom_meetings.py
```diff
@@ -134,9 +132,11 @@ def get_zoom_course(self, course: canvasapi.course.Course) -> None:
                     logger.info("Could not find a form to launch this zoom page, skipping")
                     break
 
-                self.zoom_courses.append({'account_id': course.account_id,
-                                          'course_id': course.id,
-                                          'course_name': course.name})
+                self.zoom_courses.append({'id': course.id,
+                                          'account_id': course.account_id,
+                                          'course_name': course.name,
+                                          'placement_type_id':  PlacementTypes.ZOOM
```
ssciolla, line 138:
```suggestion
                                          'placement_type_id': PlacementTypes.ZOOM
```

db/migrations/0018_add_lti_usage.py
ssciolla: Maybe `id` should stay `course_id`? I feel that's going to become confusing, since that's the convention (or at least my convention) for auto-incremented PKs, and alone it does not uniquely identify the `lit_placement` entity.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -13,8 +13,9 @@
 import requests
 from bs4 import BeautifulSoup as bs
 
+from db.db_creator import DBCreator
 from environ import ENV, DATA_DIR
-from vocab import ValidDataSourceName
+from vocab import ValidDataSourceName, PlacementTypes
```
ssciolla, line 18:
```suggestion
from vocab import PlacementTypes, ValidDataSourceName
```

online_meetings/canvas_zoom_meetings.py
```diff
@@ -201,12 +203,23 @@ def main() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.Timestamp]]]:
...
 
     zoom_courses_df = pd.DataFrame(zoom_placements.zoom_courses)
-    zoom_courses_df.index.name = "id"
+    zoom_courses_df = zoom_courses_df.set_index(["id", "placement_type_id"])
+
     zoom_courses_meetings_df = pd.DataFrame(zoom_placements.zoom_courses_meetings)
     zoom_courses_meetings_df.index.name = "id"
 
     zoom_courses_df.to_csv(os.path.join(DATA_DIR, "zoom_courses.csv"))
     zoom_courses_meetings_df.to_csv(os.path.join(DATA_DIR, "zoom_courses_meetings.csv"))
```
ssciolla, line 212: I'd like to suggest nesting this under an `if CREATE_CSVS`, like so:

https://github.com/tl-its-umich-edu/course-inventory/blob/163704da5037b81998e858f9c23225ba8fe7a9e0/course_inventory/inventory.py#L330-L334

online_meetings/canvas_zoom_meetings.py
```diff
@@ -102,13 +103,10 @@ def get_zoom_details(self, url: str, data: Dict[str, str], course_id: int):
                         self.zoom_courses_meetings.append({
                             'course_id': course_id,
                             'meeting_id': meeting['meetingId'],
-                            'meeting_number': meeting['meetingNumber'],
                             'host_id': meeting['hostId'],
                             'topic': meeting['topic'],
```
ssciolla, line 107: Do I either of these need to be dropped as well? I have not kept up on what is been approved for storing/needed by IQ. Looking back at issue #100, it looks like at least `topic` was supposed to be dropped, but things may have changed.

vocab.py
```diff
@@ -26,3 +26,7 @@ class ValidDataSourceName(Enum):
     KALTURA_API = auto()
     UNIZIN_DATA_PLATFORM_EVENTS = auto()
     UNIZIN_DATA_WAREHOUSE = auto()
+
+
+class PlacementTypes(IntEnum):
```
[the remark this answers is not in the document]
jonespm: Okay! Not sure why I love plurals so much.

online_meetings/canvas_zoom_meetings.py
jonespm: Incorporated this

online_meetings/canvas_zoom_meetings.py
jonespm: Thanks for double checking, yeah they wanted the topic removed. The host id is an alphanumeric id (rather than something that can be identified) and I think this was fine to leave in there.

db/migrations/0018_add_lti_usage.py
jonespm: My idea was that `id` in `lti_placement` would just be `course_id`. This + `placement_type_id` is unique. However you're right, without a separate auto incremented id you're right it would have to have a foreign composite key once more were added. However looking at the documentation I can't just have something like

`FOREIGN KEY (course_id, 1) REFERENCES lti_placement(id, placement_type_id) ON DELETE CASCADE ON UPDATE CASCADE`

It actually would need to be a column in `lti_zoom_meeting` to work. I switched this around only because I couldn't figure out an *easy* way of keeping the auto incremented id linked since the process.

- Pandas `to_sql` inserts all of the values into `lti_placement`
- Pandas `to_sql` Then inserts all of the values into `lti_zoom_meeting`

I'd probably have to keep a map in the code, or do a lookup against the dataframe? I guess that would work but it's more code complexity. I also figured it was better to leave the auto increment to pandas rather than SQL but that's going to cause problems when/if we upsert instead of deleting and inserting. 

1) Add a column in LTI zoom meeting (`with DEFAULT 1`) and do a composite key
2) Add an auto increment field to `lti_placement`, have course_id separate. Create some process to linkup the `lti_zoom_meeting` id to this auto increment id. I tried to think of the best way to do this and didn't quickly while still using `to_sql`. I'll think about this, maybe I can use a merge. 

Is this something you're already doing?

online_meetings/canvas_zoom_meetings.py
jonespm: Okay, didn't notice that value, thanks!

db/migrations/0018_add_lti_usage.py
ssciolla: Yeah, you've described the issue well. I wish there was some kind of unique identifier created by Canvas for an installed LTI app. Are you aware of any?

I am doing something like this for `job_run` and `data_source_status`, but it's pretty basic. It does things one at a time, and those never will need to be updated:

https://github.com/tl-its-umich-edu/course-inventory/blob/163704da5037b81998e858f9c23225ba8fe7a9e0/run_jobs.py#L45

ssciolla: In the URL for an LTI, I am seeing an external tool ID: `courses/{course_id}/external_tools/{external_tool_id}`. I'm guessing that's for LTI app, not the placement. Maybe we could cheat and just smush these together?

jonespm: Yeah, that's a good idea! That will require some code refactor but I think it could work for both cases. I had an idea for delete/insert that would work if we generated the id's in the code but not if we switched to upsert so using something like this makes more sense.

jonespm: Okay, so the external_tool_id isn't actually unique per course. It's the same for Zoom in every course. course + this id is unique but that really doesn't matter since this is static.

I think I'm going to have to auto increment in this code and optimize it later.

jonespm: I resolved this on the latest commit. I had to force push as there were conflicts with another recent change.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -99,22 +100,19 @@ def get_zoom_details(self, url: str, data: Dict[str, str], course_id: int):
...
-                            'course_id': course_id,
+                            'lti_placement_id': lti_placement_id,
                             'meeting_id': meeting['meetingId'],
-                            'meeting_number': meeting['meetingNumber'],
                             'host_id': meeting['hostId'],
-                            'topic': meeting['topic'],
-                            'join_url': meeting['joinUrl'],
                             'start_time': meeting['startTime'],
                             'status': meeting['status'],
-                            'timezone': meeting['timezone']
```
lsloan, line 110: I know #100 says many of these columns shouldn't be included in the report.  Couldn't the report just not use this data?  If there's some other good use for this data, it could be kept.

vocab.py
```diff
@@ -130,3 +130,7 @@ def copy(self) -> Dict[str, Union[ValidDataSourceName, datetime]]:
             'data_source_name': self._data_source_name,
             'data_updated_at': self._data_updated_at
         }
+
+
+class PlacementType(IntEnum):
+    ZOOM = 0
```
ssciolla, line 136: Wondering now if this should be `LTIPlacementType` or `LtiPlacementType` ...

Nit: This could also be moved up with the other enums to improve organization. Also, the `external_tool_id` might be a better ID for these types; I think that could be discussed if or when more LTI tools are reported on.

db/migrations/0020_add_lti_usage.py
```diff
@@ -0,0 +1,30 @@
...
+
+steps = [
+    step('''
+        CREATE TABLE IF NOT EXISTS lti_placement (
+            id BIGINT NOT NULL,
+            course_id INTEGER NOT NULL,
+            account_id INTEGER NOT NULL,
+            course_name VARCHAR(200),
+            placement_type_id INTEGER NOT NULL,
+            PRIMARY KEY (id, placement_type_id)
```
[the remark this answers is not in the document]
jonespm: Yeah, we only really need id now since it's basically an autoincrement. I think I just forgot about updating this. Thx.

vocab.py
jonespm: Yeah, naming things is hard! I can put it first in the file though.

jonespm: Also maybe calling it placement isn't even accurate. It's really an internal id for this external tool. Without the table it kind of loses external context.

online_meetings/canvas_zoom_meetings.py
jonespm: We're removing this data completely is removed because of a discussion with IA on a recent call where they said (and I'm paraphasing): If you aren't actively using PII stored data you shouldn't be storing it, otherwise it's a security concern.

Join URL, Topic and Meeting number were all considered to be PII for Zoom. timezone was just all the same for everyone. I was just retaining the minimal data useful for generating the report.

db/migrations/0020_add_lti_usage.py
```diff
@@ -0,0 +1,30 @@
...
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
+    '''),
+    step('''
+        CREATE TABLE IF NOT EXISTS lti_zoom_meeting (
+            id BIGINT NOT NULL,
+            lti_placement_id BIGINT NOT NULL,
+            meeting_id VARCHAR(100) NOT NULL,
+            host_id VARCHAR(100) NOT NULL,
+            start_time DATETIME NOT NULL,
+            status INTEGER,
```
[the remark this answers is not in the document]
jonespm: Yeah, really all this is useful for is generating the counts. The UUID meeting could be linked into something the Zoom team is doing but it's not useful outside of that and it isn't PII. I'm just removing based on the past conversation here.

online_meetings/canvas_zoom_meetings.py
```diff
@@ -200,12 +203,24 @@ def main() -> Sequence[DataSourceStatus]:
...
+        zoom_courses_meetings_df.to_csv(os.path.join(DATA_DIR, "zoom_courses_meetings.csv"))
+
+    # For now until this process is improved just remove all the previous records
+    logger.info('Emptying Canvas LTI data tables in DB')
+    db_creator.drop_records(
+        ['lti_placement', 'lti_zoom_meeting']
+    )
+
+    zoom_courses_df.to_sql("lti_placement", db_creator.engine, if_exists="append", index=True)
+    zoom_courses_meetings_df.to_sql("lti_zoom_meeting", db_creator.engine, if_exists="append", index=True)
```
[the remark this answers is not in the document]
jonespm: Ok I'm all for more logging!

vocab.py
jonespm: I'm just going to leave this name for now, it might get changed in a future commit anyway when I think more about it.

ssciolla: Yeah, you're right, `Placement` itself is confusing. I'm okay keeping it as is for now if you want to continue thinking about it.

online_meetings/canvas_zoom_meetings.py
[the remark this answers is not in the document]
jonespm: We could *probably* clean up this logger/to_csv/to_sql code and make this a function somewhere. Or a method in a Mixin that both of these classes extend. This also might make it easier when the upsert comes.

I guess I'll just hardcode it here for now.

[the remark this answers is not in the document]
ssciolla: Yes, what we're doing in `course_inventory` for logging is very repetitive. There's an already issue for that (#149) and that could be expanded.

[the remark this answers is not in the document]
jonespm: Just adding this logging is really making me want to refactor, so many duplicate patterns here. I also noticed the variables still referenced the old file name patterns. I'll update these.

Issue #159, opened 2020-05-14
Fix length of `course.sis_id`


ssciolla:
Today we encountered errors in local testing related to values that are too long in the `course.sis_id` field of the database. This field right now has a limit of 15. We should make it much longer to ensure that possible database insertion errors are avoided.

This will be needed as part of a bugfix release, so the migration number will have to be 15 (the next after the release). The other migration numbers should be increaed as part of a separate subsequent commit or PR, so the one migration can be cherry-picked.

ssciolla:
While this fix is an improvement, it seems that the lack of this isn't actually causing problems in production yet, because the values for this field are being truncated upon insertion. This may be because the databases are not using Strict Mode (https://dev.mysql.com/doc/refman/5.7/en/sql-mode.html). More research is needed to determine whether this is accurate and whether we need to request changes from the database administrators.

PR #160, opened 2020-05-14
Extend course.sis_id, modify post-2020.01.01 migration numbers (#159)

ssciolla:
This PR extends the `VARCHAR` length of `course.sis_id` in the database to 256 (from 15) in order to handle irregular course SIS IDs that might break the production process. This change was introduced by a new `0015` migration; subsequent migrations at or higher than that number were re-named accordingly. This PR will not be squashed, so that the first commit can be cherry-picked to the `release` branch for a bugfix release. This PR aims to resolve issue #159.


pushyamig:
Passes Test. Changes correctly reflected in DB

PR #158, opened 2020-05-14
Move custom formatting configuration to environ.py (#157)

ssciolla:
This PR reinstates a lost custom logging format by moving a call of `basicConfig` from `run_jobs.py` to `environ.py`. The PR aims to resolve issue #157.

Issue #157, opened 2020-05-14
Fix custom logging format


ssciolla:
Before `run_jobs.py` was implemented, we had a custom logging format configured for the `inventory.py` job. It looks like the migration of that to `run_jobs.py` with the simultaneous creation of `environ.py` led to this being lost.

This should be fixed if we move the `basicConfig` call in `run_jobs.py` to `environ.py`.

https://github.com/tl-its-umich-edu/course-inventory/blob/4853fe5e74d7d98b8d88e2835fb1ee5fe525e4ea/run_jobs.py#L19-L22

lsloan:
I like the formatting the way it is currently, but I'd like to see the addition of the function name.  Looks like adding `funcName` to the format string could do that.  See: https://docs.python.org/3/library/logging.html#logrecord-attributes

zqian:
verified the log message format change on the test cron output.

PR #163, opened 2020-05-15
Remove extra space in custom log format (#157)

ssciolla:
This PR removes an extra space that I missed in PR #158 and finishes resolving issue #157.

Edit: I removed used of `localLogger` from `mivideo/mivideo_extract.py` per @lsloan's request.

### DIFF
--- a/environ.py
+++ b/environ.py
@@ -30,7 +30,7 @@ with open(os.path.join(ROOT_DIR, 'config', 'env_schema.hjson')) as schema_file:
 LOG_LEVEL: str = ENV.get('LOG_LEVEL', 'INFO')
 logging.basicConfig(
     level=LOG_LEVEL,
-    format='%(asctime)s - %(name)s - %(funcName)s - %(levelname)s  - %(message)s'
+    format='%(asctime)s - %(name)s - %(funcName)s - %(levelname)s - %(message)s'
 )
 
 # Override ENV key-value pairs with values from os.environ if set
--- a/mivideo/mivideo_extract.py
+++ b/mivideo/mivideo_extract.py
@@ -130,15 +130,13 @@ class MiVideoExtract:
 
         tableName: str = 'mivideo_media_started_hourly'
 
-        localLogger = logging.getLogger(f'{logger.name}.mediaStartedHourly')
-
-        localLogger.info('Starting procedure...')
+        logger.info('Starting procedure...')
 
         lastTime: datetime = self._readTableLastTime(
             tableName, 'event_time_utc_latest', self.defaultLastTimestamp
         )
 
-        localLogger.debug('Running query...')
+        logger.debug('Running query...')
 
         dfCourseEvents: pd.DataFrame = udpDb.query(
             queries.COURSE_EVENTS, job_config=bigquery.QueryJobConfig(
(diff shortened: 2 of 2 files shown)


ssciolla:
This is why it's probably not a good idea to merge things at 4:15 on a Friday : (

environ.py
```diff
@@ -30,7 +30,7 @@
 LOG_LEVEL: str = ENV.get('LOG_LEVEL', 'INFO')
 logging.basicConfig(
     level=LOG_LEVEL,
-    format='%(asctime)s - %(name)s - %(funcName)s - %(levelname)s  - %(message)s'
+    format='%(asctime)s - %(name)s - %(funcName)s - %(levelname)s - %(message)s'
```
lsloan, line 33: I'd commented in #157 that I liked the log format that had been used at the time and I'd be happy with the addition of `funcName`. (Surprising appearance of camelCase, no?)  However, now that I've seen the log output, in this format, I like the delimiters with dash and spaces.  (Although, not **_too many_** spaces. 😉)

I also like the addition of the time.  I didn't think about it as I run the program locally, but having the timetstamp is pretty important for troubleshooting while running it on a server.  I don't really like the comma separator for fractional seconds, but it's probably overkill to try to replace `asctime` with something else.

After I've seen the results, I think the format should include `module`.  Putting it along with `funcname` (`%(funcName)s.%(funcName)s`) would be good.

lsloan: I wonder why logging doesn't offer a way to include the class name.  It seems that `module` will be the name of the file without `.py`, but I didn't see anything else in http://docs.python.org/3/library/logging.html#logrecord-attributes that mentions class.

jonespm: It looks like the class name is not available directly. So you'd either have to do some work to inspect the stack which would have a performance penalty or else use a Mixin that saves the class name [like this example. ](https://stackoverflow.com/a/35177483/3708872)

I don't think it's super important here since the funcName is here. 

Also [interesting history about the logging module](https://stackoverflow.com/a/22993896/3708872). It was originally based on Java (Log4j) and wasn't in the standard library originally. 

Also I think it's interesting that we almost always do
`logger = logging.getLogger(__name__)` which should actually be `LOGGER` since it's a often a global.

ssciolla: Yeah, I've wondered about why we (I) don't use `LOGGER`. I think I was just following what I'd seen before in T&L code; prior to starting here, I hadn't used `logging` that much.

ssciolla: @lsloan, do you really want me to add `module`? Aren't we already getting the full import path in `name`, including the module? I agree it would be nice to have `class` be supported.

ssciolla: I suppose when it's `__main__`, we don't see `run_jobs.py`, but in most cases, it'll show `mivideo.mivideo_extract`, or `course_inventory.inventory`. 

We could consider replacing `name` (which is really the logger name) with `module`, but we might lose a little information.

lsloan: If you think `module` is sufficiently represented as part of `name`, then I guess it's not necessary.  Without trying it myself, I'm not totally certain how much difference there is between the two variables.

lsloan: WRT class name, it's probably not worth worrying about.  Instead, we should ensure that multiple classes in the same module don't contain methods with the same names.

ssciolla: Okay, I'll check to make sure `module` would truly be redundant. If not, I'll merge.

ssciolla: Besides the case of `__main__` (where it would be nice to see `run_jobs`), I think the current configuration is better, as it includes the packages as well (I'm not seeing a super clear way otherwise to do this with the available attributes). I'm going to merge as is.

Issue #21, opened 2020-03-24
Retry mechanism when fetching course Published date


pushyamig:
For getting the published date, with sequential fetching of data takes a long time so we opted for the Asynchronous call approach, but sometimes we have seen canvas throws error for certain courses or we might encounter some internal code errors as well
`500  {"errors":[{"message":"An error occurred.","error_code":"internal_server_error"}],"error_report_id":7024627224}`

We could fix this by some sort of retry mechanism and with async calls.

pushyamig:
@zqian @jennlove-um The work that I put in for the fetching published date seems to be temporary and canvas seems to be releasing a new API with all the course info probably next week `“/api/v1/audit/account/courses/:account_id” `. 
I don't know if we want to implement this issue. The approach we take using that API will be totally different then. Thoughts??

ssciolla:
@pushyamig, we can close this, right?

pushyamig:
YEs

PR #164, opened 2020-05-18
I156 better published date

### LINKED ISSUE #156: Optimize the Published date process
Canvas is having problems with Audit Log API and due to that fetching the published date is taking time. Here is the proposed solution

1. Try to get the courses published date only for the courses that don’t have then and not just the whole set
2. Add a retry Mechanism if error encountered for attempts that is configurable
3. add a Timeout

pushyamig:
Fixes #156 , #21 
This PR addresses the 
1. retry logic when the canvas is sending 500/504 and 
   1. Retrying for each course (failed with HTTP error or JSON error) 3 times in case of error 
encounted. I have never seen any API error due to JSON parsing.
   2. Average Success response time is 0-30 sec. The gateway timeout error takes about 1minutes. That's the max amount of the time that is taking
2. Only fetch those published dates with new courses added as part of API call and that from the database don't have a date
   Test case went through:
       -  With an empty database
       -  Some dates already in DB ( with both null and not null published date) and few more to 
           fetch as part of new courses
 

I tried to apply a timeout to request and response. But Not effective. Especially with a response timeout will result in total failure in fetching responses. Futures-requests responses are Yield and try/catch can stop the further pulling more responses if a particular response time out

This API is not very reliable, even after applying the steps like retry/paginating/fetching dates that aren't there in DB canvas sometimes don't send data.  if I go to the browser and do the API call I can see the date.

### DIFF
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -262,7 +262,7 @@ def run_course_inventory() -> Sequence[DataSourceStatus]:
     logger.info(f"Size of courses with available workflow state: {course_available_df.shape}")
 
     course_copy_df = course_df.copy(deep=True)
-    logger.info(f"Size of courses data from API routine:  {course_copy_df.shape}")
+    logger.info(f"Size of courses data from API routine: {course_copy_df.shape}")
     logger.info(f"""Size of Published courses from API:
                 {course_copy_df[(course_copy_df['workflow_state'] == 'available')].shape}""")
     course_from_db_df = get_course_info_from_DB(db_creator_obj)
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -262,55 +262,53 @@ def run_course_inventory() -> Sequence[DataSourceStatus]:
     logger.info(f"Size of courses when workflow state available {course_available_df.shape}")
 
     course_copy_df = course_df.copy(deep=True)
-    logger.info(f"Size of courses from API  {course_copy_df.shape}")
-    logger.info(f"Size of courses from API with workflow state available  {course_copy_df[(course_copy_df['workflow_state'] == 'available')].shape}")
-
+    logger.info(f"Size of courses data from API routine:  {course_copy_df.shape}")
+    logger.info(f"""Size of Published courses from API:
+                {course_copy_df[(course_copy_df['workflow_state'] == 'available')].shape}""")
     course_from_db_df = get_course_info_from_DB(db_creator_obj)
-
-    logger.info(f"Size of courses set from DB with workflow state available {course_from_db_df.shape}")
+    logger.info(f"Size of Published courses from DB with:  {course_from_db_df.shape}")
+    published_date_in_db = course_from_db_df[(course_from_db_df['published_at'].notnull())].shape
+    logger.info(f"Size of Published courses from DB with published date: {published_date_in_db}")
     course_copy_df = pd.merge(course_copy_df, course_from_db_df, on='canvas_id', how='left')
-    logger.info(f"Size of course_copy_df after merging from Database data {course_copy_df.shape}")
+    logger.info(f"Size of course data after merging from DB data {course_copy_df.shape}")
 
     course_available_pub_date_null_df = course_copy_df.loc[(course_copy_df['workflow_state'] == 'available') &
                                                            (course_copy_df['published_at'].isnull())].copy(deep=True)
     course_ids_available_with_no_pub_date = course_available_pub_date_null_df['canvas_id'].to_list()
-    logger.info(f"published date are going to be fetched for {len(course_ids_available_with_no_pub_date)}")
+    logger.info(f"Published dates going to be fetched are: {len(course_ids_available_with_no_pub_date)}")
 
     if len(course_ids_available_with_no_pub_date) > 0:
         logger.info("*** Fetching the published date ***")
         published_dates = FetchPublishedDate(CANVAS_URL, CANVAS_TOKEN, NUM_ASYNC_WORKERS,
                                              course_ids_available_with_no_pub_date, MAX_REQ_ATTEMPTS)
+        start_time = time.time()
         published_course_date = published_dates.get_published_course_date(course_ids_available_with_no_pub_date)
-        logger.info("###########################################")
-        logger.info(f"Size Published dates fetched: {len(published_course_date)}")
+        delta = time.time() - start_time
+        str_time = time.strftime("%H:%M:%S", time.gmtime(delta))
+        logger.info(f"Published date Process took {str_time}")
+
+        newly_published_date_size = len(published_course_date)
+        logger.info(f"Size Published dates fetched from API: {newly_published_date_size}")
+        logger.info(f"Database should have {published_date_in_db[0] + newly_published_date_size} published dates ")
         if len(published_course_date) > 0:
-            pd.set_option('display.max_columns', None)
-            pd.set_option('display.max_rows', None)
             course_published_date_df = pd.DataFrame(published_course_date.items(), columns=['canvas_id', 'published_at'])
             course_published_date_df['published_at'] = pd.to_datetime(course_published_date_df['published_at'],
                                                         format=CANVAS_DATETIME_FORMAT,
                                                         errors='coerce')
-            logger.info(course_published_date_df)
             course_copy_df = pd.merge(course_copy_df, course_published_date_df, on='canvas_id', how='left')
-            logger.info(course_copy_df)
             if course_from_db_df.empty:
                 course_copy_df['published_at'] = course_copy_df['published_at_y']
             else:
                 course_copy_df['published_at'] = course_copy_df['published_at_x'].fillna(course_copy_df['published_at_y'])
             course_copy_df = course_copy_df.drop(['published_at_x', 'published_at_y'], axis=1)
-            course_final_df = course_copy_df[['canvas_id','published_at']]
-            logger.info(f"GRABBED: {course_final_df}")
-            course_df = pd.merge(course_df, course_final_df, on='canvas_id', how='left')
-            logger.info(course_df)
+            course_copy_df = course_copy_df[['canvas_id','published_at']]
+            course_df = pd.merge(course_df, course_copy_df, on='canvas_id', how='left')
         else:
             logger.info(f"No more published date fetched than what is stored in DB")
             course_df = pd.merge(course_df, course_from_db_df, on='canvas_id', how='left')
     else:
         logger.info(f"All courses seems to have Published dates, so fetching Published dates routine not needed")
-    #
-    logger.info("*** Checking for courses available and no published date ***")
-    logger.info(course_df[(course_df['workflow_state'] == 'available') & (course_df['published_at'].isnull())])
-    #
+        logger.info(f"Database should have {published_date_in_db[0]} published dates ")
     course_df['created_at'] = pd.to_datetime(course_df['created_at'],
                                              format=CANVAS_DATETIME_FORMAT,
                                              errors='coerce')
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -259,7 +259,7 @@ def run_course_inventory() -> Sequence[DataSourceStatus]:
     # Gather course data
     course_df = gather_course_data_from_api(ACCOUNT_ID, TERM_IDS)
     course_available_df = course_df.loc[course_df.workflow_state == 'available'].copy(deep=True)
-    logger.info(f"Size of courses when workflow state available {course_available_df.shape}")
+    logger.info(f"Size of courses with available workflow state: {course_available_df.shape}")
 
     course_copy_df = course_df.copy(deep=True)
     logger.info(f"Size of courses data from API routine:  {course_copy_df.shape}")
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -266,7 +266,7 @@ def run_course_inventory() -> Sequence[DataSourceStatus]:
     logger.info(f"""Size of Published courses from API:
                 {course_copy_df[(course_copy_df['workflow_state'] == 'available')].shape}""")
     course_from_db_df = get_course_info_from_DB(db_creator_obj)
-    logger.info(f"Size of Published courses from DB with:  {course_from_db_df.shape}")
+    logger.info(f"Size of published courses from DB with: {course_from_db_df.shape}")
     published_date_in_db = course_from_db_df[(course_from_db_df['published_at'].notnull())].shape
     logger.info(f"Size of Published courses from DB with published date: {published_date_in_db}")
     course_copy_df = pd.merge(course_copy_df, course_from_db_df, on='canvas_id', how='left')
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -228,18 +228,17 @@ def pull_sis_section_data_from_udw(section_ids: Sequence[int], conn: connection)
     return udw_section_df
 
 
-def get_course_info_from_db(db_creator_obj: DBCreator) -> pd.DataFrame:
+def get_pub_course_info_from_db(db_creator_obj: DBCreator) -> pd.DataFrame:
     logger.info(f"Getting the course info from {db_creator_obj.db_name} database")
     course_from_db_df = pd.read_sql(f'''select canvas_id, published_at from course 
                             where workflow_state = 'available' ;''',
                                     db_creator_obj.engine)
-    logger.info(course_from_db_df.head())
     return course_from_db_df
 
 # Entry point for run_jobs.py
 
 
-def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.Timestamp]]]:
+def run_course_inventory() -> Sequence[DataSourceStatus]:
     logger.info("* run_course_inventory")
     # Initialize DBCreator object
     db_creator_obj = DBCreator(INVENTORY_DB)
@@ -255,7 +254,7 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
     logger.info(f"Size of courses with available workflow state: {course_available_df.shape}")
 
     course_copy_df = course_df.copy(deep=True)
-    course_from_db_df = get_course_info_from_db(db_creator_obj)
+    course_from_db_df = get_pub_course_info_from_db(db_creator_obj)
 
     fetch_publish_date = FetchPublishedDate(CANVAS_URL, CANVAS_TOKEN, NUM_ASYNC_WORKERS,
                                             course_copy_df, course_from_db_df, MAX_REQ_ATTEMPTS)
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -228,21 +228,13 @@ def pull_sis_section_data_from_udw(section_ids: Sequence[int], conn: connection)
     return udw_section_df
 
 
-def get_course_info_from_DB(db_creator_obj: DBCreator) -> pd.DataFrame:
+def get_course_info_from_db(db_creator_obj: DBCreator) -> pd.DataFrame:
     logger.info("getting the course info from Database")
     course_from_db_df = pd.read_sql(f'''select canvas_id, published_at from course 
                             where workflow_state = 'available' ;''',
-                            db_creator_obj.engine)
+                                    db_creator_obj.engine)
     logger.info(course_from_db_df.head())
     return course_from_db_df
-
-
-def get_course_from_db_without_pub_date(df) -> List[int]:
-    course_without_pub_date_df = df[df['published_at'].isnull()]
-    course_id_with_no_pub = course_without_pub_date_df['canvas_id'].tolist()
-    logger.info(f"Courses list from Database with out published date{len(course_id_with_no_pub)}: {course_id_with_no_pub}")
-    return course_id_with_no_pub
-
 
 # Entry point for run_jobs.py
 
@@ -265,7 +257,7 @@ def run_course_inventory() -> Sequence[DataSourceStatus]:
     logger.info(f"Size of courses data from API routine: {course_copy_df.shape}")
     logger.info(f"""Size of Published courses from API:
                 {course_copy_df[(course_copy_df['workflow_state'] == 'available')].shape}""")
-    course_from_db_df = get_course_info_from_DB(db_creator_obj)
+    course_from_db_df = get_course_info_from_db(db_creator_obj)
     logger.info(f"Size of published courses from DB with: {course_from_db_df.shape}")
     published_date_in_db = course_from_db_df[(course_from_db_df['published_at'].notnull())].shape
     logger.info(f"Size of published courses from DB with published date: {published_date_in_db}")
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -255,53 +255,17 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
     logger.info(f"Size of courses with available workflow state: {course_available_df.shape}")
 
     course_copy_df = course_df.copy(deep=True)
-    logger.info(f"Size of courses data from API routine: {course_copy_df.shape}")
-    logger.info(f"""Size of Published courses from API:
-                {course_copy_df[(course_copy_df['workflow_state'] == 'available')].shape}""")
     course_from_db_df = get_course_info_from_db(db_creator_obj)
-    logger.info(f"Size of published courses from DB with: {course_from_db_df.shape}")
-    published_date_in_db = course_from_db_df[(course_from_db_df['published_at'].notnull())].shape
-    logger.info(f"Size of published courses from DB with published date: {published_date_in_db}")
-    course_copy_df = pd.merge(course_copy_df, course_from_db_df, on='canvas_id', how='left')
-    logger.info(f"Size of course data after merging from DB data: {course_copy_df.shape}")
 
-    course_available_pub_date_null_df = course_copy_df.loc[(course_copy_df['workflow_state'] == 'available') &
-                                                           (course_copy_df['published_at'].isnull())].copy(deep=True)
-    course_ids_available_with_no_pub_date = course_available_pub_date_null_df['canvas_id'].to_list()
-    logger.info(f"Published dates going to be fetched are: {len(course_ids_available_with_no_pub_date)}")
+    fetch_publish_date = FetchPublishedDate(CANVAS_URL, CANVAS_TOKEN, NUM_ASYNC_WORKERS,
+                                            course_copy_df, course_from_db_df, MAX_REQ_ATTEMPTS)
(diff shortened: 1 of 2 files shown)


course_inventory/inventory.py
```diff
@@ -353,8 +409,6 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
         canvas_course_usage_df.to_csv(os.path.join('data', 'canvas_course_usage.csv'), index=False)
         logger.info('Wrote data to data/canvas_course_usage.csv')
```
ssciolla, line 411: Please remove this newline so there is only one left here.

course_inventory/published_date.py
```diff
@@ -33,32 +39,49 @@ def get_next_page_url(self, response):
...
+        course_id = int(response.result().url.split('?')[0].split('/')[-1])
         if 'next' in results:
             url_ = results['next']['url']
-            self.published_course_next_page_list.append(url_)
-            logger.info(f"Pagination size {len(self.published_course_next_page_list)} for published_at date")
+            logger.debug('Fetching the next page URL')
+            # This update or add to the retry bucket
+            self.published_date_retry_bucket[course_id] = {
+                    'url': url_,
+                    'count': 0,
```
ssciolla, line 49: Unindent once?

course_inventory/published_date.py
```diff
@@ -33,32 +39,49 @@ def get_next_page_url(self, response):
             logging.debug('The api call do not have Link headers')
             return None
 
+        course_id = int(response.result().url.split('?')[0].split('/')[-1])
```
ssciolla, line 42: Do you think there is any performance cost to calling `result` twice (once here, once above)? Also, this seems a little fragile, endpoint-specific and not very generaliz-able. Is there another way we can get the `course_id`? Could it be passed as a parameter?

ssciolla: I guess you're also using this pattern below. If you don't want to change this, that's fine, just thought I would put this idea out there.

course_inventory/published_date.py
```diff
@@ -1,6 +1,7 @@
 import logging
 import time
 import json
+import requests
```
ssciolla, line 4: Doesn't look like `time` or `requests` are used anymore.

course_inventory/inventory.py
```diff
@@ -319,6 +374,7 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
         'data_updated_at': udw_update_datetime
     }
 
+
```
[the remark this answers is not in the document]
pushyamig: Done

course_inventory/inventory.py
```diff
@@ -291,6 +345,7 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
     enroll_delta = time.time() - enroll_start
     logger.info(f'Duration of process (seconds): {enroll_delta}')
 
+
```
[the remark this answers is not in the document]
pushyamig: Done

course_inventory/inventory.py
```diff
@@ -235,10 +235,28 @@ def pull_sis_section_data_from_udw(section_ids: Sequence[int], conn: connection)
     return udw_section_df
 
 
+def get_course_info_from_DB(db_creator_obj: DBCreator) -> pd.DataFrame:
```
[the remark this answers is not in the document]
pushyamig: Done

course_inventory/published_date.py
pushyamig: Done

course_inventory/published_date.py
pushyamig: I took `response.result()` as a variable,  I don't know how much overhead it will add for calling over and over. But this seems to be a good idea to do.
Since the response is Async I can't pass the course id and the best way to get the course_id is from the Response object. I have tested the pattern of URL that is sent with both regular and paginated URL seems to work

course_inventory/published_date.py
```diff
@@ -11,13 +12,18 @@
...
         self.canvas_token = canvas_token
         self.canvas_ids = canvas_ids
+        self.retry_attempts = retry_attempts
         self.num_workers = num_workers
         self.published_course_date = {}
-        self.published_course_next_page_list = []
+        #{332371: {'url': 'https://umich.instructure.com/api/v1/audit/course/courses/332371?page=bookmark:WzE1Nzg1Mjan
+        #           &per_page=100', 'count': 0},
+        # 343033: {'url': 'https://umich.instructure.com/api/v1/audit/course/courses/343033?per_page=100', 'count': 1},
+        # 343043: {'url': 'https://umich.instructure.com/api/v1/audit/course/courses/343043?per_page=100', 'count': 1}}
```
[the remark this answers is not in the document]
pushyamig: Added Type hinting and in this case, it makes sense to do that. 
I also added a simple example for understanding of the way the data is structured than getting a clue from type hint.

course_inventory/published_date.py
pushyamig: Removed it

course_inventory/published_date.py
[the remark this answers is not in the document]
ssciolla: Sounds good, thanks

course_inventory/published_date.py
ssciolla: Yeah, I think it makes sense to do what you did and only call `result()` once. Thanks

course_inventory/inventory.py
```diff
@@ -247,27 +265,63 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
...
-    course_available_ids = course_available_df['canvas_id'].to_list()
-    published_dates = FetchPublishedDate(CANVAS_URL, CANVAS_TOKEN, NUM_ASYNC_WORKERS, course_available_ids)
-    published_course_date = published_dates.get_published_course_date(course_available_ids)
-    course_published_date_df = pd.DataFrame(published_course_date.items(), columns=['canvas_id', 'published_at'])
-    course_df = pd.merge(course_df, course_published_date_df, on='canvas_id', how='left')
-
-    logger.info("*** Checking for courses available and no published date ***")
-    logger.info(course_df[(course_df['workflow_state'] == 'available') & (course_df['published_at'].isnull())])
-    
+    course_available_df = course_df.loc[course_df.workflow_state == 'available'].copy(deep=True)
```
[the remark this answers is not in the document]
pushyamig: I have moved the logic to `FetchPublishedDate` class

course_inventory/inventory.py
```diff
@@ -247,27 +265,63 @@ def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.T
...
+    logger.info(f"""Size of Published courses from API:
+                {course_copy_df[(course_copy_df['workflow_state'] == 'available')].shape}""")
+    course_from_db_df = get_course_info_from_DB(db_creator_obj)
+    logger.info(f"Size of Published courses from DB with:  {course_from_db_df.shape}")
+    published_date_in_db = course_from_db_df[(course_from_db_df['published_at'].notnull())].shape
+    logger.info(f"Size of Published courses from DB with published date: {published_date_in_db}")
+    course_copy_df = pd.merge(course_copy_df, course_from_db_df, on='canvas_id', how='left')
+    logger.info(f"Size of course data after merging from DB data {course_copy_df.shape}")
+
+    course_available_pub_date_null_df = course_copy_df.loc[(course_copy_df['workflow_state'] == 'available') &
```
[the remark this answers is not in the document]
pushyamig: I have moved this code to `FetchPublishedDate` code

course_inventory/inventory.py
```diff
@@ -234,10 +234,21 @@ def pull_sis_section_data_from_udw(section_ids: Sequence[int], conn: connection)
...
+                            where workflow_state = 'available' ;''',
+                                    db_creator_obj.engine)
+    logger.info(course_from_db_df.head())
+    return course_from_db_df
+
 # Entry point for run_jobs.py
 
-def run_course_inventory() -> Sequence[DataSourceStatus]:
+
+def run_course_inventory() -> Sequence[Dict[str, Union[ValidDataSourceName, pd.Timestamp]]]:
```
ssciolla, line 248: Not sure what happened here, but what was there before was correct. Also, I'd like to have two newlines above the comment and underneath the new function you added.

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
+                                                    on='canvas_id', how='left')
+            if is_published_date_all_empty:
+                courses_with_pub_date_col_df['published_at'] = courses_with_pub_date_col_df['published_at_y']
+            else:
+                courses_with_pub_date_col_df['published_at'] = courses_with_pub_date_col_df['published_at_x'].fillna(
+                    courses_with_pub_date_col_df['published_at_y'])
+
+            courses_with_pub_date_col_df = courses_with_pub_date_col_df.drop(['published_at_x', 'published_at_y'],
+                                                                             axis=1)
+        logger.info(f"Size of **Newly Published dates fetched from API: {len(self.published_course_date)}")
```
ssciolla, line 214:
```suggestion
        logger.info(f"Size of newly published dates fetched from API: {len(self.published_course_date)}")
```

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
+        logger.info(
+            f"Database should have {published_date_in_db[0] + len(self.published_course_date)} published dates")
```
ssciolla, line 216:
```suggestion
            f"There are now {published_date_in_db[0] + len(self.published_course_date)} published dates")
```

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
+        logger.info(f"Published dates going to be fetched are: {len(course_avail_with_no_pub_date_list)}")
+        if len(course_avail_with_no_pub_date_list) == 0:
+            courses_with_pub_date_col_df = courses_with_pub_date_col_df[['canvas_id', 'published_at']]
+            logger.info("No more published date to fetch than what is stored in DB")
+            logger.info(f"Database should have {published_date_in_db[0]} published dates")
+            return courses_with_pub_date_col_df
+
+        self.get_published_course_date(course_avail_with_no_pub_date_list)
+
+        if len(self.published_course_date) > 0:
```
ssciolla, line 197: Maybe we could add an `else` saying that no new data was collected? I like to start my `if/else` blocks with the fail case so that a short `else` isn't hidden after a bunch of code. So something like this:

```
if len(self.published_course_date) == 0:
   logger.info('No new published dates for courses were collected')
else:
   ... (the rest of your code)
```

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
+        logger.info(f"""Size of Published courses from API: {state_available_courses_from_api}""")
+        logger.info(f"Size of published courses from DB with: {self.course_data_from_db.shape}")
+        published_date_in_db = self.course_data_from_db[(self.course_data_from_db['published_at'].notnull())].shape
+        logger.info(f"Size of published courses from DB with published date: {published_date_in_db}")
+        course_with_pub_date_added_from_df = pd.merge(self.course_data_from_api, self.course_data_from_db,
+                                                      on='canvas_id', how='left')
+        logger.info(f"Size of course data after merging with DB data: {course_with_pub_date_added_from_df.shape}")
+        return course_with_pub_date_added_from_df
+
+    def get_published_course_date(self, course_ids, retry_list=None) -> None:
```
ssciolla, line 152: Nit-pick: The arguments here could use type hints for clarity. `retry_list` is confusing since I think the outermost data type is `Dict`?

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
-        if len(self.published_course_next_page_list) != 0:
-            logger.info(f"""Pagination size {len(self.published_course_next_page_list)} with published_at date items
-                         {self.published_course_next_page_list}""")
-            self.get_published_course_date(course_ids, self.published_course_next_page_list)
-
-        return self.published_course_date
+        if len(self.published_date_retry_bucket) != 0:
+            logger.info(f"""Retrying now with list size {len(self.published_date_retry_bucket)} 
+                        {self.published_date_retry_bucket}""")
+            self.get_published_course_date(course_ids, self.published_date_retry_bucket)
```
ssciolla, line 175: Do we need to pass `course_ids` here? It seems like all the information that is used during extra runs is contained within the `retry_bucket`. Maybe we could make the `course_ids` parameter have a default value of `None`? That might require some additional logic.

course_inventory/published_date.py
ssciolla: You are logging the size of `published_course_date` later, so maybe it's not necessary?

course_inventory/published_date.py
```diff
@@ -28,37 +40,56 @@ def get_next_page_url(self, response):
...
+        logger.info(f"Parsing the response for Course: {course_id}")
+        logger.info(f"Pagination info {course_id} {response.result().links}")
+        logger.info(f"Time taken to get the response for {course_id} : {response.result().elapsed}")
         status = response.result().status_code
         published_date_found = False
         if status != 200:
-            logger.info(f"Response not successful with status code {status} due to {response.result().text}")
+            logger.info(
+                f"""Response unsuccessful for {course_id} status: {status} and time taken:{response.result().elapsed}  
+                        due to {response.result().text}""")
```
ssciolla, line 84:
```suggestion
            logger.warning(f"Request was unsuccessful for {course_id}")
            logger.warning(f"Response status: {status}; time taken: {response.result().elapsed}; response text: {response.result().text}")
```

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
-        logger.debug(f"Parsing the published date took {str_time} ")
         return
 
-    def get_published_course_date(self, course_ids, next_page_links=None):
-        logger.info("Starting of get_published_course_date call")
-        with FuturesSession(max_workers=self.num_workers) as session:
+    def retry_logic_with_error(self, course_id, url) -> None:
+        if course_id in self.published_date_retry_bucket:
+            retry_count = self.published_date_retry_bucket[course_id]['count']
+            if retry_count > self.retry_attempts - 1:
```
ssciolla, line 122: Nit-pick: I think you could make this `if retry_count >= self.retry_attempts` if you didn't want to subtract. You might not even need the `>` sign...

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
+                # Don't want to retry more
+                logger.info(f"Reached max retry attempts for {course_id} with count {retry_count}, Removing... ")
+                self.published_date_retry_bucket.pop(course_id)
+            else:
+                # We will give one more chance to retry
+                logger.info(f"Retry more for {course_id}  ")
```
ssciolla, line 128:
```suggestion
                logger.info(f"Another retry attempt will be made for {course_id}")
```

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
+                self.published_date_retry_bucket[course_id]['count'] += 1
+        else:
+            logger.info(f"Putting course {course_id} to retry list")
```
ssciolla, line 131:
```suggestion
            logger.info(f"Adding course {course_id} to retry bucket")
```

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
+            self.published_date_retry_bucket[course_id] = {
+                'url': url,
+                'count': 1,
+            }
+        logger.info(f"Retry list size {len(self.published_date_retry_bucket)} for published_at date")
```
ssciolla, line 136: Not sure if this log message is needed, or if it should go somewhere else.

course_inventory/inventory.py
```diff
@@ -234,10 +234,21 @@ def pull_sis_section_data_from_udw(section_ids: Sequence[int], conn: connection)
     return udw_section_df
 
 
+def get_course_info_from_db(db_creator_obj: DBCreator) -> pd.DataFrame:
+    logger.info("getting the course info from Database")
+    course_from_db_df = pd.read_sql(f'''select canvas_id, published_at from course 
+                            where workflow_state = 'available' ;''',
+                                    db_creator_obj.engine)
+    logger.info(course_from_db_df.head())
```
ssciolla, line 242: Let's make this `DEBUG` level?

course_inventory/inventory.py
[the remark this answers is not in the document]
ssciolla: We definitely should fix the type hints here.

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
+
+    def filter_courses_to_fetch_published_date(self) -> pd.DataFrame:
+        pd.set_option('display.max_column', None)
+        logger.info(f"Size of courses data from API routine: {self.course_data_from_api.shape}")
+        state_available_courses_from_api = self.course_data_from_api[
+            (self.course_data_from_api['workflow_state'] == 'available')].shape
+        logger.info(f"""Size of Published courses from API: {state_available_courses_from_api}""")
+        logger.info(f"Size of published courses from DB with: {self.course_data_from_db.shape}")
```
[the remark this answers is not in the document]
pushyamig: I am just getting the courses that are available from DB and not unpublished courses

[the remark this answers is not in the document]
ssciolla: Ah, you are correct. Maybe we need to change the `get_course_info_from_db` function to be called `get_pub_course_info_from_db`?

course_inventory/published_date.py
pushyamig: I want to keep this for tracking purposes how many courses are being added to retry list. This should be ok

course_inventory/inventory.py
pushyamig: I am going to delete this

course_inventory/published_date.py
```diff
@@ -1,25 +1,37 @@
...
         self.canvas_token = canvas_token
-        self.canvas_ids = canvas_ids
+        self.course_data_from_api = course_data_from_api
+        self.course_data_from_db = course_data_from_db
+        self.retry_attempts = retry_attempts
         self.num_workers = num_workers
         self.published_course_date = {}
-        self.published_course_next_page_list = []
+        # { course_id: {'url': 'https://instructure.com','count': 0} }
+        self.published_date_retry_bucket: Dict[int, Dict[str, Any]] = {}
```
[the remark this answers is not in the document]
pushyamig: I have added type hints

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
+        else:
+            logger.info(f"Putting course {course_id} to retry list")
+            self.published_date_retry_bucket[course_id] = {
+                'url': url,
+                'count': 1,
+            }
+        logger.info(f"Retry list size {len(self.published_date_retry_bucket)} for published_at date")
+
+    def filter_courses_to_fetch_published_date(self) -> pd.DataFrame:
+        pd.set_option('display.max_column', None)
```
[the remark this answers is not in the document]
pushyamig: I left then by mistake I was using to see the whole Dataframe as logging was chopping off. I will remove then

course_inventory/inventory.py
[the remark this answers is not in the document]
pushyamig: I did not add those type hits, it was part of the master code I pulled in. What should be the value?

course_inventory/published_date.py
pushyamig: Will make the change

course_inventory/inventory.py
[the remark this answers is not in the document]
ssciolla: @pushyamig, what you have is not what is in `master`: https://github.com/tl-its-umich-edu/course-inventory/blob/700b986abf7eacacd46d8421da860dbc4561386a/course_inventory/inventory.py#L231-L236

course_inventory/published_date.py
```diff
@@ -28,37 +40,56 @@ def get_next_page_url(self, response):
...
             return
 
-        start_time = time.time()
-
         logger.info(f"published courses date collected so far : {len(self.published_course_date)}")
+        url = response.result().url
+        course_id = int(url.split('?')[0].split('/')[-1])
+
+        logger.info(f"Parsing the response for Course: {course_id}")
+        logger.info(f"Pagination info {course_id} {response.result().links}")
```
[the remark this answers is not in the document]
pushyamig: Make sense, I will change it

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
...
+            self.published_date_retry_bucket[course_id] = {
+                'url': url,
+                'count': 1,
+            }
+        logger.info(f"Retry list size {len(self.published_date_retry_bucket)} for published_at date")
+
+    def filter_courses_to_fetch_published_date(self) -> pd.DataFrame:
+        pd.set_option('display.max_column', None)
+        logger.info(f"Size of courses data from API routine: {self.course_data_from_api.shape}")
+        state_available_courses_from_api = self.course_data_from_api[
```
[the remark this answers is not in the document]
pushyamig: bad name, I changed to `available_courses_from_api`

course_inventory/published_date.py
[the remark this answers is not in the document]
pushyamig: I changed the name.

course_inventory/published_date.py
pushyamig: yeah... exactly

course_inventory/published_date.py
pushyamig: I don't want to change anything with this. This should be fine

ssciolla: I think it seems silly to require this as a parameter if it's not used. You're right that it's working fine, but it doesn't seem like a best practice.

course_inventory/published_date.py
```diff
@@ -75,40 +106,113 @@ def published_date_resp_parsing(self, response):
                 published_date_found = True
                 self.published_course_date.update({course_id: event['created_at']})
                 logger.info(f"Published Date {event['created_at']} for course {course_id}")
+                if course_id in self.published_date_retry_bucket:
+                    logger.info(f"Going to remove {course_id} from retry list {self.published_date_retry_bucket}")
+                    self.published_date_retry_bucket.pop(course_id)
+                    logger.info(f"Removed {course_id} removed from retry list {self.published_date_retry_bucket}")
```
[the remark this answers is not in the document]
pushyamig: I am changing this to `len(self.published_date_retry_bucket)` instead of the whole dump of retry obj

course_inventory/published_date.py
pushyamig: I have added the type hint

course_inventory/published_date.py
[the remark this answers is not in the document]
pushyamig: Done

course_inventory/published_date.py
pushyamig: It's fine for now

course_inventory/inventory.py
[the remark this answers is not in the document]
pushyamig: Done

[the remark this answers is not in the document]
pushyamig: I am not sure why I changed it

course_inventory/published_date.py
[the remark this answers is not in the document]
pushyamig: Done

course_inventory/published_date.py
pushyamig: Done

course_inventory/inventory.py
pushyamig: Done

course_inventory/published_date.py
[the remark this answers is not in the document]
pushyamig: Done

course_inventory/published_date.py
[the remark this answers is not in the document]
pushyamig: Done

course_inventory/published_date.py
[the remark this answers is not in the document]
pushyamig: Done

course_inventory/published_date.py
pushyamig: Done

course_inventory/published_date.py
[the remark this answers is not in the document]
pushyamig: Done

pushyamig:
@ssciolla I have resolved most of the what you have suggested, a couple I did not agree. This should be working now

course_inventory/inventory.py
```diff
@@ -228,10 +228,20 @@ def pull_sis_section_data_from_udw(section_ids: Sequence[int], conn: connection)
     return udw_section_df
 
 
+def get_pub_course_info_from_db(db_creator_obj: DBCreator) -> pd.DataFrame:
+    logger.info(f"Getting the course info from {db_creator_obj.db_name} database")
+    course_from_db_df = pd.read_sql(f'''select canvas_id, published_at from course 
+                            where workflow_state = 'available' ;''',
+                                    db_creator_obj.engine)
+    return course_from_db_df
+
```
ssciolla, line 237: I would add a newline here.

course_inventory/inventory.py
```diff
@@ -228,10 +228,20 @@ def pull_sis_section_data_from_udw(section_ids: Sequence[int], conn: connection)
...
 # Entry point for run_jobs.py
 
+
```
ssciolla, line 240: And remove this one here.

course_inventory/published_date.py
```diff
@@ -75,40 +106,111 @@ def published_date_resp_parsing(self, response):
...
+        logger.info(f"Size of published courses from API: {available_courses_from_api}")
+        logger.info(f"Size of published courses from DB with: {self.course_data_from_db.shape}")
+        published_date_in_db = self.course_data_from_db[(self.course_data_from_db['published_at'].notnull())].shape
+        logger.info(f"Size of published courses from DB with published date: {published_date_in_db}")
+        course_with_pub_date_added_from_df = pd.merge(self.course_data_from_api, self.course_data_from_db,
+                                                      on='canvas_id', how='left')
+        logger.info(f"Size of course data after merging with DB data: {course_with_pub_date_added_from_df.shape}")
+        return course_with_pub_date_added_from_df
+
+    def get_published_course_date(self, course_ids, retry_list: Dict[int, str] = None) -> None:
```
ssciolla, line 151: I still find the name `retry_list` confusing since the type of the argument is a `Dict`. Also, I think this type hint is wrong: it should probably be `Union[None, Dict[str, Dict[str, Any]]]`, like `self.published_date_retry_bucket`.

Issue #165, opened 2020-05-18
All SIS ID data types should be VARCHAR(256)


ssciolla:
In some local testing, @pushyamig found that sometimes the `account.sis_id` can sometimes contain trailing dashes and hyphens. While these are infrequent, the data type should either be flexible enough to handle this, or there should be an additional function that validates this field and sets invalid values to `None`.

Edit: After further discussion, we decided that to keep things simple, all SIS IDs (for `course`, `course_section`, and `account`) should use the same data type of `VARCHAR(256)`, as the Canvas data model specifies them as `varchar`. This leniency will prevent errors and leave custom validation logic out of the code, while still allowing joins on valid IDs to the SIS dataset.

zqian:
Verified the sis_id in account, course and course_section table are of VARCHAR type. It is still of int type in term table. I will update the issue title to reflect this.

PR #168, opened 2020-05-19
Relax data types of SIS IDs (#165)

ssciolla:
To resolve insertion errors and create consistency, this PR changes via a migration two SIS ID fields (`course_section.sis_id` and `account.sis_id`) to the `VARCHAR(256)` data type in the database, mirroring the Canvas data model. (Note that the `course.sis_id` was previously changed to this type in PR #160.) The PR also removes a mapping function previously used to normalize `course_section.sis_id`, as well as an unnecessary string conversion call in the `slim_down_course_data` function (the data was already a string). The PR aims to resolve issue #165.


ssciolla:
Thanks, @jonespm, for the speedy review.

PR #170, opened 2020-05-20
Support PTVSD debugging in the container

### LINKED ISSUE #169: Support PTVSD debugging in the container
Now that the job is running in the container, we'll have to do remote debugging to debug this in VSCode. I think all we'd have to do is add the ptvsd module to the requirements and have a way to change the run command to 

`-m ptvsd --host localhost --port 5678 --wait run_jobs.py`

Maybe by using a custom shell script to start in debug verses non-debug? 

Python code could also set this up verses using the CLI syntax.

jonespm:
Fixes #169

Previous startup remains the same, you need to use a different
docker-compose to add in PTVSD changes.


jonespm:
The only change to the current code is adding a new package to the requirements. Everything else is new/optional. 

I like this method way of overriding the docker-compose.

pushyamig:
This works beautifully!

lsloan:
Thanks for inviting me to review this PR, @jonespm.  However, I don't use VS Code or have it installed, so I don't have any way to test this.

I had looked at the changes before you requested my review.  I would have suggested removing comments from the JSON, but I see you've done that in a commit subsequent to my viewing.

So, I'll recuse myself from this review and leave it in the hands of the others.

requirements.txt
```diff
@@ -13,3 +13,4 @@ SQLAlchemy==1.3.16
 strict-rfc3339==0.7
 -e git+https://github.com/tl-its-umich-edu/api-utils-python@v1.3#egg=umich-api
 yoyo-migrations==7.0.2
+ptvsd>=4.3.2,<4.3.99
```
ssciolla, line 16: With all other dependencies we've pinned to the exact bugfix release. Was the range here intentional? Should we be using ranges elsewhere?

docker-compose-ptvsd.yaml
```diff
@@ -0,0 +1,7 @@
+version: '3'
```
ssciolla, line 1: Any consideration to running this from a directory instead? I think this is a cool idea, but it would be nice not to add yet another file to the root level.

requirements.txt
ssciolla: Also, I will politely request that this be put in the proper place to maintain alphabetical order.

jonespm: Yeah, I just copied from MyLA, I can lock the version.

They haven't released a new version of this package since August 2019 but if this isn't updated together when VSCode is updated, generally has broken compatibility with VSCode. But we can update that when it happens.

pushyamig: I think the range for this library should be fine here since this is purely for local development and we want to update the library automatically to support a new version of VSCode  released.

jonespm:
Sure @lsloan, I just invited everyone. Do you have any way you're doing remote debugging with whatever software you're using? I find that remote debugging really speeds up development time and it even works when this is running out on Openshift. I don't think I could work in Java without using it in Eclipse and just working on the last issue I was thinking it would have saved me time if I just added this so I could inspect the variables.

requirements.txt
ssciolla: I know we've gone back and forth about this with our projects. I found this fairly lucid StackOverflow answer that I thought might be worth sharing: https://stackoverflow.com/a/44938662

It seems they're suggesting using ranges in PyPi packages and pinning dependencies for custom applications.

docker-compose-ptvsd.yaml
jonespm: I'm not sure what you mean. We could probably move all of the build files (Dockerfile, docker-compose*) into a subdirectory if you wanted to clean up the root.

ssciolla: I mean, yeah, that's exactly what I mean. I feel there might be some value in having `Dockerfile` and `docker-compose.yaml` in the root for transparency (but maybe I could be convinced otherwise), but having this extra `docker-compose-ptvsd.yaml` only for use with specific technology at the root level ... I dunno. I just wanted to know if you'd considered this, I'm not asking for the change. Could it go in `.vscode`?

requirements.txt
jonespm: I'd personally prefer we have every version locked, including the versions of packages installed in apt. This should make the builds consistently reproducible forever.

But the problem is that it requires effort to have a consistent release schedule that also includes reviewing and updating all of the dependencies. (Or like the SO says have a service that reminds us to do that)

I've seen as a project gets bigger with less focus, dependency updating usually is forgotten. :( It would be nice to improve on that, but it also hasn't caused any problems yet.

docker-compose-ptvsd.yaml
jonespm: Sure this file could go anywhere, it just needs to be on the command line. I feel like the biggest negative of a ton of files on the root is when viewing on Github you have to scroll to see the README like with [Open Edx](https://github.com/edx/edx-platform) or [Canvas](https://github.com/instructure/canvas-lms). Wow Canvas has like 10 docker-compose* files on the root! :)

requirements.txt
ssciolla: Yeah, I think I agree. I think having someone do research on the best services for this would be great, would help us for many projects. In the meantime, you just have to do a little legwork, and I still think we should be doing this for each minor or major release cycle. Note that I did do for 2020.01.01 (in PR #140). It doesn't take that long, especially when we only have 15ish dependencies.

docker-compose-ptvsd.yaml
jonespm: I feel like this thread is bikeshedding as @johnpj would say. ;)

ssciolla: Yeah, it also bugs me in VSCode, since I like to right click on open space in the left-pane directory tree to create a new directory, and when there are a million root-level files (MyLA comes to mind), there isn't much space to work with.

ssciolla: I mean, fine, I'll take the put down in stride. I like things to be organized, that's just part of who I am.

jonespm: I wouldn't consider it a "put down", I just think if we want to clean up the root it should be on a different issue rather than one-off. At least half of the files could be moved. I mostly use the terminal is VSCode to do a lot of stuff so I've never right clicked in the left pane, but different people work differently!

jonespm: Created at #171. I'll move this file for now to end this thread and move on to another issue.

Issue #174, opened 2020-05-21
Canvas API may return duplicate courses


ssciolla:
Today, the two jobs in OpenShift failed because of duplicate `course` primary keys, for different courses. This appears to be unexpected API behavior, as I don't believe this has happened before. Still, we should likely modify the existing code to `drop_duplicates` and note how many courses were dropped, so we can avoid this error while also monitoring API anomalies.

ssciolla:
I discussed the problem with @jonespm, as this may also occur with his use of `canvasapi` in the `CANVAS_ZOOM_MEETINGS` job. Ultimately we'll want to avoid fetching the same data multiple times (see issue #162), but we'll need to handle this there as part of resolving this issue in the short-term.

jonespm:
It was mentioned on #176 that we might want to move the code from get_lti_tabs out one more level and only call this once.

jonespm:
This one is not yet finished. We could probably bump it with #162 to the next release but I think it could be finished by itself tomorrow. I've never seen any case where there were duplicate courses returned as part of this process.

zqian:
verified the function in place with the current cron log file:

2020-06-02 14:31:57,970 - course_inventory.inventory - gather_course_data_from_api - INFO - Course records with students: x
2020-06-02 14:31:57,970 - course_inventory.inventory - gather_course_data_from_api - INFO - Dropped x course record(s) with no students
2020-06-02 14:31:58,016 - course_inventory.inventory - gather_course_data_from_api - INFO - Dropped x duplicate course record(s)

jonespm:
I believe because of chances of this happening are low and the entire process should be changed in #162 I'm going to close this for now. Doing this would require:

- Retrieving all courses in a term
- Removing Duplicates (if any)
- Add in ADD_COURSE_IDS courses that aren't already in the list
- Process all these courses

This should all be refactored into one place to populate the courses table so it can be used by future classes.

PR #175, opened 2020-05-21
Handle possibility of duplicate courses being inserted to `course` table (#174)

ssciolla:
Similar to PR #146, this PR adds duplicate dropping to the creation of `course` records to avoid possible errors when inserting into the database. The modifications include the use of `pd.DataFrame.drop_duplicates` with the `subset` parameter to limit duplicate consideration to the PK (`canvas_id`) and the `keep` parameter to ensure the second record found is kept. This change is responding to incidents in the production and test processes where a different course in each process seems to have been duplicated. The PR aims to resolve issue #174.

### DIFF
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -177,6 +177,10 @@ def gather_course_data_from_api(account_id: int, term_ids: Sequence[int]) -> pd.
 
     course_df = pd.DataFrame(course_dicts_with_students)
     course_df = course_df.drop(['total_students'], axis='columns')
+    orig_course_count = len(course_df)
+    course_df = course_df.drop_duplicates(subset=['canvas_id'], keep='last')
+    logger.info(f'{orig_course_count - len(course_df)} course records were dropped')
+
     logger.debug(course_df.head())
     return course_df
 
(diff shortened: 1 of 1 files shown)


ssciolla:
If you want to test this, you can do something like this (from a testing branch): 
https://github.com/ssciolla/course-inventory/blob/f448a2f77feed0b353c8d570302f5b089ca37dfc/course_inventory/inventory.py#L178-L188

course_inventory/inventory.py
```diff
@@ -177,6 +177,10 @@ def gather_course_data_from_api(account_id: int, term_ids: Sequence[int]) -> pd.
 
     course_df = pd.DataFrame(course_dicts_with_students)
     course_df = course_df.drop(['total_students'], axis='columns')
+    orig_course_count = len(course_df)
+    course_df = course_df.drop_duplicates(subset=['canvas_id'], keep='last')
```
ssciolla, line 181: Does `last` seem right in this case? Not sure it matters too much...

zqian: either last or first should be fine.

I wonder whether the duplicate course could exist in course_dicts (line 167) already? Should the check duplicate id be done at that level?

ssciolla: I mean, either way the duplicates would be dropped, right? It's easier to deal with duplicates with `pandas`; I suppose I could create the dataframe earlier and do the filtering with `pandas` also. Were you concerned about something in particular?

zqian: Yes, the duplicates would be dropped in the end anyway. Please go ahead with the current impl.

Issue #161, opened 2020-05-15
Refine error case of make_request_using_api_utils


ssciolla:
_Raised by @pushyamig during review of PR #151_

Currently, the `make_request_using_api_utils` function always returns a `Response` object, even if the request failed three times. Though this hasn't ever happened in my memory, we should revisit the return type of this function and whether or not we want the program to continue after failing in this way.

zqian:
Please see the latest handling in issue #184

PR #177, opened 2020-05-26
Exit if max attempt limit is reached in make_request_using_api_utils (#161)

ssciolla:
This PR modifies the `make_request_using_api_utils` function from `inventory.py` -- used by the `course`, `account`, and `term` Canvas data fetching -- so that it exits the program if the maximum number of attempts are reached. This ensures that the `Response` object returned by the function is always a valid one. The PR aims to resolve issue #161.

I thought for a little while about whether the program should continue after failing multiple times, and I decided that for these requests, the dataset would be incomplete without the data. We would either be missing 100 courses or a foreign key constraint would fail from `course` -> `account` or `term`. Note that I have not yet seen this situation in our deployed processes; if we are concerned that this might happen, we can bump the max request attempts to `4`.

### DIFF
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -1,5 +1,5 @@
 # standard libraries
-import json, logging, os, time
+import json, logging, os, sys, time
 from json.decoder import JSONDecodeError
 from typing import Any, Dict, List, Sequence, Union
 
@@ -70,7 +70,8 @@ def make_request_using_api_utils(url: str, params: Union[Dict[str, Any], None] =
                 logger.info('Beginning next attempt')
 
     logger.error('The maximum number of request attempts was reached')
-    return response
+    logger.error('Data could not be gathered; the program will exit')
+    sys.exit(1)
 
 
 def gather_term_data_from_api(account_id: int, term_ids: Sequence[int]) -> pd.DataFrame:
--- a/course_inventory/inventory.py
+++ b/course_inventory/inventory.py
@@ -70,7 +70,8 @@ def make_request_using_api_utils(url: str, params: Union[Dict[str, Any], None] =
                 logger.info('Beginning next attempt')
 
     logger.error('The maximum number of request attempts was reached')
-    logger.error('Data could not be gathered; the program will exit')
+    logger.error(f'Data could not be gathered from the URL with the ending "{url}"')
+    logger.error('The program will exit')
     sys.exit(1)
 
 


pushyamig:
@ssciolla your rationale seems for exiting if retry fails seems ok. I have seen different API errors out in different ways. You might experience JSON parsing error might not be true for another API. I have also seen that for certain API retry sometime don't fetch the data you want when tried momentarily. 

This is OK.

course_inventory/inventory.py
```diff
@@ -70,7 +70,8 @@ def make_request_using_api_utils(url: str, params: Union[Dict[str, Any], None] =
                 logger.info('Beginning next attempt')
 
     logger.error('The maximum number of request attempts was reached')
-    return response
+    logger.error('Data could not be gathered; the program will exit')
```
pushyamig, line 73: will it make sense to have the URL output in the log message?

ssciolla: The URL for the failed requests? Sure, yeah, I could add that.

pushyamig: yes

ssciolla: Okay, I think this was resolved. See this commit: https://github.com/tl-its-umich-edu/course-inventory/pull/177/commits/695d45f6e17b24988dfae84826f5d2fb4032473c

course_inventory/inventory.py
```diff
@@ -70,7 +70,9 @@ def make_request_using_api_utils(url: str, params: Union[Dict[str, Any], None] =
                 logger.info('Beginning next attempt')
 
     logger.error('The maximum number of request attempts was reached')
-    return response
+    logger.error(f'Data could not be gathered from the URL with the ending "{url}"')
+    logger.error('The program will exit')
```
zqian, line 74: Will the whole program/cron exit, or just this single URL call exit? Please update the log message accordingly.

ssciolla: The whole program will exit (thus I think the log message is accurate). I explained my reasoning in the PR description; @pushyamig seems to concur.

I suppose we might want other jobs (`MIVIDEO`, or `CANVAS_ZOOM_MEETINGS`) to continue running, but that would require a bunch more logic and modification of other functions. Keep in mind this is a failure case, and I have never seen it happen spontaneously.

zqian: OK. Hope we won't encounter this failure.

jonespm: Possible the better thing to do if we just wanted to abort this job would be to raise a custom exception, catch it up in run_jobs. Then the other jobs could continue when this one fails. 

Jenkins would still consider this script a failure but it wouldn't block anything else. I'm fine if you want to consider this for the future.

course_inventory/inventory.py
```diff
@@ -70,7 +70,9 @@ def make_request_using_api_utils(url: str, params: Union[Dict[str, Any], None] =
...
+    sys.exit(1)
```
jonespm, line 75: There are already 2 sys.exit's in the code. It might make it easier to debug if we were actually exiting with a different code at each one. Maybe we can have a constant defined somewhere like in vocab.EXIT_CODES?

Cant exit with any code between 1-127 to indicate a failure.

jonespm:
I'm okay with the changes, I just left some comments about how this could be improved, but they're not necessary to implement now.

course_inventory/inventory.py
ssciolla: I think this is a good idea. I'll create a separate issue.

course_inventory/inventory.py
ssciolla: Not a bad idea, but aren't the log error messages prior to the exit code just going to be more informative?

PR #181, opened 2020-05-26
added Apache 2 LICENSE and NOTICE files

zqian:


zqian:
> Is there any harm in specifying a file type ending (.txt or .md)? I think it would be okay with or without, but I prefer to specify this when I can.

With the [Apache recommendation](https://infra.apache.org/apply-license.html), both LICENSE and NOTICE files shall be used. Same approach used in MyLA and Student Explorer project.

ssciolla:
> With the [Apache recommendation](https://infra.apache.org/apply-license.html), both LICENSE and NOTICE files shall be used. Same approach used in MyLA and Student Explorer project.

Ah yes, thanks for linking to that. They specifically say that they prefer if you don't but there's nothing stopping you from doing so. Interesting.

jonespm:
Sure if you go through the Github License workflow this is what it adds. the NOTICE is optional. Looks good.

Issue #180, opened 2020-05-26
Update .dockerignore to ensure HJSON files are excluded, use proper syntax


ssciolla:
As part of PR #134, I neglected to update the `.dockerignore` file, which tells Docker which files to disregard when building images. While this isn't a huge risk, it is possible for developers who are using a mixed approach of `virtualenv` and `Docker` to include credential files in images that they build locally if they are keeping a copy of their `env.hjson` in `config/secrets`. To eliminate that risk, we should update the patterns in `.dockerignore` to cover the new HJSON file ending.

https://docs.docker.com/engine/reference/builder/#dockerignore-file

ssciolla:
While looking into this, I also discovered that the `.dockerignore` syntax works a little differently than `.gitignore` syntax. Previously I had expected them to be the same. To find files with certain patterns recursively in all child directories, you need to use the special Docker `**/` syntax.

zqian:
To test:

- comment out the :/app line under services->job -> volumns within docker-compose.yaml
- docker-compose build; docker-compose up -d; docker exec course_inventory /bin/bash
- verify those files are not in the docker image, e.g. no env.hjson inside config folder, etc.

PR #182, opened 2020-05-26
Update .dockerignore (#180)

ssciolla:
This PR updates the repo's `.dockerignore` file to ensure unnecessary and sensitive files are not included in Docker images built locally. Changes include use of the `**/` Docker-specific syntax to properly find files with certain patterns in any directory; widening of the `env` pattern to capture HJSON configuration files; and the removal of the `.vscode` directory from the list (as it might be needed in our deployment environments). The PR aims to resolve issue #180.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,6 +1,7 @@
 # config
 **/env.*
 venv
+.vscode
 
 # data
 **/*.csv
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,13 +1,12 @@
 # config
-env.*
+**/env.*
 venv
 
 # data
-*.csv
+**/*.csv
 .data
 
 # etc
 .DS_Store
-.vscode
-*.pyc
-__pycache__
+**/*.pyc
+**/__pycache__
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,5 @@
 # config
-env.json
+env.*
 venv
 
 # data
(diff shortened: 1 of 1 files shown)


.dockerignore
```diff
@@ -1,13 +1,12 @@
...
+**/*.csv
 .data
 
 # etc
 .DS_Store
-*.pyc
-.vscode
-__pycache__
\ No newline at end of file
+**/*.pyc
```
jonespm, line 11: So the .dockerignore has a different syntax from .gitignore? That's kind of annoying, thanks for noticing this!

.dockerignore
```diff
@@ -1,13 +1,12 @@
...
 
 # data
-*.csv
+**/*.csv
 .data
 
 # etc
 .DS_Store
-*.pyc
-.vscode
```
jonespm, line 12: I don't think there's any need to include this in the image but there's no harm in it either. You'd have to be running vscode inside the container for this to matter.

.dockerignore
ssciolla: Yeah, it's described near the bottom of this section: https://docs.docker.com/engine/reference/builder/#dockerignore-file

.dockerignore
ssciolla: I guess I was thinking about the debugger (which I still don't have a lot of experience with). You mentioned at one point that you can use this OpenShift. Would you not need `.vscode` to do that? I can add this back if it's not needed.

ssciolla: Okay, that makes sense. I'll add it back to the file.

ssciolla:
Thanks for the reviews! I'm merging now.

Issue #184, opened 2020-05-29
Use custom exception rather than exiting so other jobs can continue


ssciolla:
Possible the better thing to do if we just wanted to abort this job would be to raise a custom exception, catch it up in run_jobs. Then the other jobs could continue when this one fails. 

Jenkins would still consider this script a failure but it wouldn't block anything else. I'm fine if you want to consider this for the future. I'm going to unresolve this as we may want to create a new issue for this.

_Originally posted by @jonespm in https://github.com/tl-its-umich-edu/course-inventory/pull/177_

zqian:
Test steps as @ssciolla suggested, tested this locally by editing "any of the Canvas request functions (I usually add an a to the end of the request URL ending for the term process). That will fail with repeated 404 errors and you’ll see the exception triggered:
Like change line 87 of inventory.py to the following: 
term_url_ending = url_ending_with_scope + str(term_id) + 'a'
"

The test is a success:

- The failed job did not stop others from running. 
- I got the expected messages in the log file:
"2020-06-05 09:53:53,515 - __main__ - run - ERROR - JobError: Data could not be gathered from the URL with the ending "aa/CanvasReadOnly/accounts/1/terms/167a"
2020-06-05 09:53:53,515 - __main__ - run - ERROR - An error prevented the COURSE_INVENTORY job from finishing
2020-06-05 09:53:53,515 - __main__ - run - INFO - The program will continue running other jobs"
- no data inserted into job_run table.

PR #185, opened 2020-05-29
Implement JobError custom exception (#184)

ssciolla:
Per @jonespm's suggestion, this PR implements a custom exception called `JobError` that is raised whenever a job needs to be cancelled before it is completed. A `try/except` will catch the exception in `run_jobs`. If this occurs, no metadata will be stored in the `job_run` and `data_source_status` database tables. Subsequent jobs will still run, though an `ERROR`-level log message will still be recorded (leading to Jenkins marking a build as a failure). The PR aims to resolve issue #184.

Resource(s) consulted:
* https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions


vocab.py
```diff
@@ -46,6 +48,21 @@ def __str__(self) -> str:
...
 
 
+# Exception(s)
+
+class Error(Exception):
+    '''Base class for exceptions for the project'''
+
+
+class JobError(Error):
+    '''Exception raised for errors that should prompt a job cancellation'''
```
lsloan, line 58: Does `JobError` need to be based on `Error`, also defined here?  Isn't there an existing Python error that could be used or just base it on `Exception` directly?

Also, isn't `Error` ambiguous?  Couldn't there be conflicts with `Error` classes from other modules used by this application?  I'm thinking it should be called `CourseInventoryBaseError`, if the class is really necessary.

jonespm: Yeah seems like `JobError` could just extend `Exception` unless we were doing more with Error. We could probably also call it `JobException(Exception):`

ssciolla: I was following what the docs (https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions) describe, since it was presented as a best practice. I can rename or remove it if you don’t think it’s important.

ssciolla: There is no base Python `Error` class that I could find, just `Exception`.

ssciolla: `SomethingError` seems to be the standard naming pattern for user-defined Exceptions, and Exceptions in general (e.g. `KeyError`).

run_jobs.py
```diff
@@ -61,14 +61,19 @@ def run(self) -> None:
...
+            self.finished_at = time.time()
+
+            delta = self.finished_at - self.started_at
+            str_time = time.strftime('%H:%M:%S', time.gmtime(delta))
+            logger.info(f'Duration of job run: {str_time}')
+
+            self.create_metadata()
+        except JobError as je:
+            logger.error(f'JobError: {je}')
+            logger.info(f'An error prevented the {self.name} job from finishing')
```
zqian, line 75: I think this line can be combined with line 74 at the error level instead, since the job has been stopped due to the error.

ssciolla: @zqian, I do not want these to be in the same log message. I see the first one (74) as just reporting the `JobError` message, while the second (75) is explaining how the program will respond to the error. If you want me to bump the level for the second to `ERROR`, I can do that, please tell me if that's what you want.

vocab.py
ssciolla: I didn't want to wait any longer for more direction, so I changed the base class to `CourseInventoryBaseError`, as @lsloan suggested.

run_jobs.py
ssciolla: I will bump this to `ERROR`, but I'd like to keep them as separate messages. I feel there's going to be too much information crammed in one line otherwise.

ssciolla:
Thank you for the reviews. I'm going to merge now.

vocab.py
lsloan: @ssciolla, I'm sorry if you were frustrated by waiting for responses.  After I gave my review for the PR, I became engrossed in some of the recent MiVideo questions and I found a bug to be corrected.  I didn't notice you'd been waiting for more feedback.

I had hoped that my approval of your PR and the way I'd worded my comments you'd understand that I didn't mean them to block the merge.  I just wanted to make sure a potential problem had enough attention.

FWIW, yes, I saw that you'd cited the exception tutorial.  It's a good resource and I'm glad you mentioned it. However, I think if everybody follows it and names their base exception `Error`, there's even more possibility for ambiguity and conflict.  I wish they would address that in the tutorial.  Either to say that base exception classes should have more explicit names or to confirm that I'm wrong about ambiguity and conflict.  (Maybe I can open an issue to request the tutorial be updated.)

Issue #162, opened 2020-05-15
Link MiVideo and Canvas Zoom Meetings up to Courses


jonespm:
Currently all of the jobs run independently, and the only script that populates courses in the courses table is inventory. The problem is that this script currently when it runs drops all of the courses and it's also possible that a different set of courses might be processed. 

We could refactor this whole thing and have

A top level job that processes though courses

* Get courses in a term that match the criteria (or specific courses if defined)
  * Runs the course inventory
* Get courses from the LTI placement report (or optionally from the terms or specific courses)
  * Runs the LTI placements

Or we could just have some method that creates courses for all of these scripts. I don't know how MiVideo process would fit into this courses job?

Mentioned in https://github.com/tl-its-umich-edu/course-inventory/pull/155#discussion_r424720313

PR #187, opened 2020-06-01
Switching from names over to ids for external tools

### LINKED ISSUE #186: Use ID's instead of names to identify if tool is one we should include
We should use external tool ids rather than names to tell if the tool is one we should include. Using a name could indicate false positives.

jonespm:
Fixes #186

The names internally are of the format
context_external_tool_##### so I'm retaining that format.


config/env_blank.hjson
```diff
@@ -25,7 +25,12 @@
         "API_CLIENT_SECRET": "",
         # Canvas API
         "CANVAS_URL": "",
-        "CANVAS_TOKEN": ""
+        "CANVAS_TOKEN": "",
+        # For processing canvas_placements
+        # ID in each object can be duplicated if the same tool is duplicated
+        "SUPPORTED_LTI_TOOLS" : {
+            "context_external_tool_12345": {"name": "Zoom", "id": 0},
+        }
```
ssciolla, line 33: This looks like it will work, but if we're going to start doing something with the `external_tool` ID, maybe it just makes sense to start using that ID instead of the `0` and `1`. 

I'd also like to try to keep the configuration as simple as possible. `context_external_tool` isn't going to change right? Maybe this could be condensed to something like the following (the key and value could also be flipped):

```
        "SUPPORTED_LTI_TOOLS" : {
            "ZOOM": 12345
        }
```

Thoughts?

jonespm: I would say it "probably" won't change but I don't know for sure. This is how it's reported in the api call. 

I debating between splitting off the numeric id and just using this numeric id instead of the full string
`id = s.rsplit('_', 1)[1]`

or just putting the full string here. 

I think having an internal id would still need to be something we have because I can imagine multiple context_tools might be the same tool we want to report on. Like 12345 and 51231 could both be Zoom. So I figured we'd use the Canvas ID as the key here.

ssciolla: If the nature of the relationship between LTI tool (a Zoom LTI) and LTI provider (Zoom the company) changes from 1:1 to M:1, aren't we going to want to record or identify the separate LTI tools for each service? I dunno, it just seems like using this ID is going to be more precise than letting the configuration classify them (and ultimately hiding differences between them). That's my thinking, but I'll defer to you on this.

If you can remove that repeated string -- or at least make it separately configurable -- I think that would help simplify things.

jonespm: Well an LTI launch is always going to be 1:1. However you can place new tools at different account levels in Canvas (or even possibly at the site level). So we could have a Zoom at the top level and also have a Zoom in LSA. Would we want to call them something different in the config? I don't know? I guess this gives us the flexibility to change the name or use the same identifiers. 

I am going to simplify the string though but I *think* it's possible we may want to just consider multiple ids in Canvas to be the same for our reporting.

config/env_schema.hjson
```diff
@@ -57,6 +57,13 @@
                 # Canvas API
                 "CANVAS_URL": {"type": "string"},
                 "CANVAS_TOKEN": {"type": "string"},
+                "SUPPORTED_LTI_TOOLS" : {
+                    type: "object"
+                    "properties": {
+                        "name": {"type": "string"}
+                        "id": {"type": "integer"}
+                    }
+        }
```
ssciolla, line 66: Nit-pick: Please fix indentation; this `{` should be under the beginning of `SUPPORTED_LTI_TOOLS`.

config/env_blank.hjson
```diff
@@ -25,7 +25,12 @@
         "API_CLIENT_SECRET": "",
         # Canvas API
         "CANVAS_URL": "",
-        "CANVAS_TOKEN": ""
+        "CANVAS_TOKEN": "",
+        # For processing canvas_placements
+        # ID in each object can be duplicated if the same tool is duplicated
+        "SUPPORTED_LTI_TOOLS" : {
```
ssciolla, line 31:
```suggestion
        "SUPPORTED_LTI_TOOLS": {
```

lti_placements/canvas_placements.py
```diff
@@ -35,14 +29,20 @@ class CanvasLtiPlacementProcessor:
...
         self.zoom_placements = ZoomPlacements(self.canvas)
+        if not supported_lti_tools:
+            supported_lti_tools = {}
+        self.supported_lti_tools = supported_lti_tools
 
     def generate_lti_course_report(self,
                                    canvas_account_id: int,
                                    enrollment_term_ids: Union[Sequence[int], None],
-                                   add_course_ids: Union[List[int], None] = None,
+                                   add_course_ids: Union[List[int], None],
```
ssciolla, line 45: Nit-pick: if you're removing the `None` default parameter, could the type hint be simplified?

config/env_schema.hjson
ssciolla: Sorry, I'd provide a suggestion but spaces and tabs are weird in the GitHub UI.

README.md
```diff
@@ -88,6 +88,7 @@ a more lenient and customizable variant of JSON. Complete the following items in
     `CANVAS` | `API_CLIENT_SECRET` | The client secret for authenticating to the API Directory.
     `CANVAS` | `CANVAS_URL` | The Canvas instance URL to be used as the base URL for API requests that use the `CANVAS TOKEN`.
     `CANVAS` | `CANVAS_TOKEN` | The Canvas token used for authenticating to the API when not using the U-M API Directory.
+    `CANVAS` | `SUPPORTED_LTI_TOOLS` | The tool definitions supported by the CANVAS_LTI job. See the env_blank for a sample of a configuration.
```
ssciolla, line 91: Not related to your PR, but can you fix the end backtick for the `kaltura_categories_full_name_in` variable below? It's using a single quote at the end now.

jonespm: Yup, fixed!

lti_placements/canvas_placements.py
jonespm: Well someone can still pass None, I just removed this default to clean up the args a little bit. I'm not sure what you mean by "type hint simplified"

config/env_schema.hjson
jonespm: Yeah, weird!

lti_placements/canvas_placements.py
```diff
@@ -245,9 +249,16 @@ def main() -> Sequence[DataSourceStatus]:
...
-    lti_processor = CanvasLtiPlacementProcessor(canvas_env.get("CANVAS_URL"), canvas_env.get("CANVAS_TOKEN"))
-    lti_processor.generate_lti_course_report(canvas_env.get("CANVAS_ACCOUNT_ID", 1), canvas_env.get(
-        "CANVAS_TERM_IDS", []), canvas_env.get("ADD_COURSE_IDS", []), True)
+    lti_processor = CanvasLtiPlacementProcessor(
+        canvas_env.get("CANVAS_URL"),
+        canvas_env.get("CANVAS_TOKEN"),
+        canvas_env.get("SUPPORTED_LTI_TOOLS", {}))
+
+    lti_processor.generate_lti_course_report(
+        canvas_env.get("CANVAS_ACCOUNT_ID", 1),
```
zqian, line 258: looks like the only the root account will be checked for LTI tool report. I think we need to do recursive loop into all published courses in all accounts.

Maybe instead of using CanvasAPI's account.get_courses, we can use the course list reported in the course_inventory job?

jonespm: Yeah that's issue #162. Would require a larger refactor.

jonespm: Also I think that the other process is using the UM API, this is going direct to Canvas. This *might* be able to use the UM API now but I'm not sure. I wanted to support UM API as a backend to CanvasAPI but that is also some work.

lti_placements/canvas_placements.py
ssciolla: Well, you are only passing `[]` now, so I guess I just don't know why we're including the possibility of `None` anymore; no one else is using this code. This isn't a big deal; you can leave it as is.

lti_placements/canvas_placements.py
ssciolla: I don't quite understand what you mean, @zqian, by the "recursive loop into all published courses." We're not doing this in `COURSE_INVENTORY`. Wouldn't all the courses we care about be under the root account `1`, even if they have a specific sub-account?

Issue #52, opened 2020-04-02
Issue with UUID duplicate dropping?


ssciolla:
Might be off-base here, but doesn't the resulting dataframe need to be re-assigned here? 

[report, lines 93-95](https://github.com/jonespm/course-inventory/blob/373ea575e5ea6830f68a022a02cd0d684a758002/online_meetings/report.py#L93-L95)

My understanding is that it doesn't mutate. See [documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.drop_duplicates.html).

ssciolla:
Oh, doh, just saw the `inplace`; I've never used that. I'm closing this.

jonespm:
It actually isn't a bad suggestion and we can switch it. It's not a bug though as you've noticed. I think this is related to #40 and can be switched then. 

I've used inplace often in the past but aside from fillna it doesn't seem like it has much of an advantage and internally it just creates a new object anyway. And they're talking about making it a FutureWarning. Some of the confusion here seems like developers don't really know if there's an advantage to using it or not. 

https://github.com/pandas-dev/pandas/issues/16529

PR #190, opened 2020-06-04
Improve MiVideo course ID parsing (#150)

### LINKED ISSUE #150: improve MiVideo category handling
As discussed with @ssciolla, it would be safer to better define how categories are used to query the Kaltura API and how to transform the appropriate categories into Canvas course ID references.

The code currently uses regular expressions to find categories that begin with a prefix (`Canvas_UMich`, by default) and search them for any number and treat that as the Canvas course ID number.  It would be better to search for the specific category formats:

* `Canvas_UMich>site>channels>𝒏𝒏𝒏𝒏`
* `Canvas_UMich>site>channels>𝒏𝒏𝒏𝒏>InContext`

Questions for the implementation:

1. Avoid regular expressions.  They're great tools, but they can sometimes match strings that weren't intended (e.g., #188).  If regular expressions must be used, reduce the amount of (or eliminate) globbing operators in them.
1. If limited regular expression globbing is used, at what point does using regular expression processing become more expensive or complicated than using simple string manipulation functions?
1. Should the regular expressions or interesting strings be moved to the application configuration file?
1. Instead of renaming the `categories` column to `course_id`, process `categories`, assign it to `course_id`, and then drop `categories`.

lsloan:
Resolves #150.

## Summary
Simplifies the transformation of Kaltura media categories to Canvas course ID numbers and removes dependency on regular expressions.  Both make it easier for maintainers to understand.

## Details
* Don't rename the `categories` column to `course_id` immediately.  Create the new column by transforming the data, then delete the old column.  Maybe not as efficient, but easier to understand.
* Avoid using Pandas' `str` property of `Series` objects.  Use `map()` instead, which is something maintainers will probably be familiar with from base Python, JavaScript, and other languages.  🎩 @ssciolla 
* Replace regular expression use with splitting category strings on `>`, the delimiter Kaltura uses in its category hierarchies.
* Probably improved efficiency in the removal of non-decimal course IDs part of the process.  No longer uses a regular expression to find and log the bad items, followed by using `str.isdecimal()` to remove them, which processes all `course_id` values **_twice_**.  Instead, process them once with `map(str.isdecimal)` to get a Boolean index of all rows and use negation (`~`) to log them.

## Review
I **_humbly_** request that comments on this PR be nested under a single review per reviewer.  It helps me when the comments are organized this way, especially if the comments request changes in the code.

If I've requested a review from you but you are definitely unable to review this PR, please leave a comment saying so (please state _why_ you're unable to review, that would be appreciated), tell the other reviewers that the review will be left to them, and remove your name from the list of reviewers.

### DIFF
--- a/mivideo/mivideo_extract.py
+++ b/mivideo/mivideo_extract.py
@@ -334,7 +334,7 @@ class MiVideoExtract:
             logger.debug(f'Course IDs to be removed:\n{courseData[invalidCourseIdIndex]}')
             courseData = courseData[validCourseIdIndex]  # keep the valid ones
 
-        courseData.drop_duplicates(inplace=True)
+        courseData = courseData.drop_duplicates()
 
         return courseData
 
--- a/mivideo/mivideo_extract.py
+++ b/mivideo/mivideo_extract.py
@@ -304,23 +304,25 @@ class MiVideoExtract:
             resultDictionaries, columns=('id', 'categories',)
         ).rename(columns={'id': 'media_id', })
 
-        # categories start to become course IDs
+        # split category CSVs to multiple new rows
         courseData = (
             courseData
-                .assign(course_id=courseData['categories'].str.split(','))
-                .explode('course_id')
-                .drop('categories', axis=1)  # 1: columns
+                .assign(categories=courseData['categories'].str.split(','))
+                .explode('categories')
         )
 
-        courseData['in_context'] = courseData['course_id'].str.endswith('>InContext')
+        courseData['in_context'] = courseData['categories'].str.endswith('>InContext')
 
-        # remove original category prefix and suffix from each course ID
-        # would like to avoid regex, but at least this is SIMPLE regex
+        # remove category prefix and suffix to get course IDs
+        # (prefer to avoid regex, but at least this doesn't use globbing)
         courseData['course_id'] = (
-            courseData['course_id'].str
+            courseData['categories'].str
                 .replace('^Canvas_UMich>site>channels>', '', regex=True)
                 .replace('>InContext$', '', regex=True)
         )
+
+        # categories have been processed, drop unneeded column
+        courseData = courseData.drop('categories', axis=1)  # 1: columns
 
         # find and drop invalid, non-decimal course IDs
         # (e.g., "Shared Repository", "5430bafe907cca901a0f11646470dd64244ebd5f", etc.)
--- a/mivideo/mivideo_extract.py
+++ b/mivideo/mivideo_extract.py
@@ -331,8 +331,7 @@ class MiVideoExtract:
         if (not validCourseIdIndex.all()):  # not all all index items are True
             logger.debug('Non-decimal course IDs to be removed:\n'
                          f'{courseData[~validCourseIdIndex]}')
-
-        courseData = courseData[validCourseIdIndex].drop_duplicates()
+            courseData = courseData[validCourseIdIndex].drop_duplicates()
 
         return courseData
 
--- a/mivideo/mivideo_extract.py
+++ b/mivideo/mivideo_extract.py
@@ -329,9 +329,12 @@ class MiVideoExtract:
         validCourseIdIndex = courseData['course_id'].map(str.isdecimal)
 
         if (not validCourseIdIndex.all()):  # not all all index items are True
-            logger.debug('Non-decimal course IDs to be removed:\n'
-                         f'{courseData[~validCourseIdIndex]}')
-            courseData = courseData[validCourseIdIndex].drop_duplicates()
+            invalidCourseIdIndex = ~validCourseIdIndex
+            logger.info(f'Removing ({invalidCourseIdIndex.values.sum()}) non-decimal course IDs...')
+            logger.debug(f'Course IDs to be removed:\n{courseData[invalidCourseIdIndex]}')
+            courseData = courseData[validCourseIdIndex]  # keep the valid ones
+
+        courseData.drop_duplicates(inplace=True)
 
         return courseData
 
--- a/mivideo/mivideo_extract.py
+++ b/mivideo/mivideo_extract.py
@@ -302,26 +302,35 @@ class MiVideoExtract:
         """
         courseData: pd.DataFrame = pd.DataFrame.from_records(
             resultDictionaries, columns=('id', 'categories',)
-        ).rename(columns={'id': 'media_id', 'categories': 'course_id', })
+        ).rename(columns={'id': 'media_id', })
 
-        courseData = courseData.assign(
-            course_id=courseData['course_id'].str.split(',')).explode('course_id')
-
-        courseData['in_context'] = [
-            c.endswith('>InContext') for c in courseData['course_id']]
-
-        courseData['course_id'] = courseData['course_id'].str.replace(
-            r'^' + categoryFilter + r'.*>([0-9]+)(>InContext)?$',
-            lambda m: m.groups()[0], regex=True
+        # categories start to become course IDs
+        courseData = (
+            courseData
+                .assign(course_id=courseData['categories'].str.split(','))
+                .explode('course_id')
+                .drop('categories', axis=1)  # 1: columns
         )
 
-        # find and drop non-decimal course IDs
-        # (e.g., like category "Canvas_UMich>site>channels>Shared Repository")
-        badCourseIds = courseData[courseData['course_id'].str.contains('[^0-9]', regex=True)]
-        if (not badCourseIds.empty):
-            logger.debug(f'Non-numeric course IDs to be removed:\n{badCourseIds}')
+        courseData['in_context'] = courseData['course_id'].str.endswith('>InContext')
 
-        courseData = courseData[courseData['course_id'].str.isdecimal()].drop_duplicates()
+        # remove original category prefix and suffix from each course ID
+        # would like to avoid regex, but at least this is SIMPLE regex
+        courseData['course_id'] = (
+            courseData['course_id'].str
+                .replace('^Canvas_UMich>site>channels>', '', regex=True)
+                .replace('>InContext$', '', regex=True)
+        )
+
+        # find and drop invalid, non-decimal course IDs
+        # (e.g., "Shared Repository", "5430bafe907cca901a0f11646470dd64244ebd5f", etc.)
+        validCourseIdIndex = courseData['course_id'].str.isdecimal()
+
+        if (not validCourseIdIndex.all()):  # not all all index items are True
+            logger.debug('Non-numeric course IDs to be removed:\n'
+                         f'{courseData[~validCourseIdIndex]}')
+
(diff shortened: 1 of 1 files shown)


mivideo/mivideo_extract.py
```diff
@@ -297,31 +296,45 @@ def _makeCourseData(resultDictionaries: Sequence[Dict], categoryFilter: str) ->
...
-            r'^' + categoryFilter + r'.*>([0-9]+)(>InContext)?$',
-            lambda m: m.groups()[0], regex=True
-        )
+        courseData['in_context'] = courseData['categories'] \
+            .map(lambda c: c.endswith('>InContext'))
+
+        # get course ID from '>'-delimited category hierarchy string
+        columnIndex = 3
+        courseData['course_id'] = \
+            courseData['categories'].map(lambda c: (c + '>' * columnIndex).split('>')[columnIndex])
```
ssciolla, line 322: Are you adding `>>>` to the end of each string you are parsing to prevent index errors? Is this a standard practice? It seems maybe error-prone to me, and like maybe it would be much clearer to use [`str.count()`](https://docs.python.org/3/library/stdtypes.html#str.count) in a condition of some sort. Something like this?
```
def parseCategory(input: str) -> Union[str, None]:
   columnIndex = 3
   sepCount = input.count(`>`)
   if sepCount >= columnIndex:
      return input.split(`>`)[columnIndex)
   else:
     logger.warning(f'Unusual category input: {input}')
     return None
```

mivideo/mivideo_extract.py
```diff
@@ -297,31 +296,45 @@ def _makeCourseData(resultDictionaries: Sequence[Dict], categoryFilter: str) ->
...
+
+        # categories have been processed, drop unneeded column
+        courseData = courseData.drop('categories', axis=1)  # 1: columns
+
+        # find and drop invalid, non-decimal course IDs
+        # (e.g., "Shared Repository", "5430bafe907cca901a0f11646470dd64244ebd5f", etc.)
+        validCourseIdIndex = courseData['course_id'].map(str.isdecimal)
```
ssciolla, line 329: Any reason you're not using `isnumeric` here? Aren't we handling integers?

mivideo/mivideo_extract.py
```diff
@@ -297,31 +296,45 @@ def _makeCourseData(resultDictionaries: Sequence[Dict], categoryFilter: str) ->
...
-        if (not badCourseIds.empty):
-            logger.debug(f'Non-numeric course IDs to be removed:\n{badCourseIds}')
+        if (not validCourseIdIndex.all()):  # not all all index items are True
+            invalidCourseIdIndex = ~validCourseIdIndex
+            logger.info(f'Removing ({invalidCourseIdIndex.values.sum()}) non-decimal course IDs...')
+            logger.debug(f'Course IDs to be removed:\n{courseData[invalidCourseIdIndex]}')
+            courseData = courseData[validCourseIdIndex]  # keep the valid ones
 
-        courseData = courseData[courseData['course_id'].str.isdecimal()].drop_duplicates()
+        courseData.drop_duplicates(inplace=True)
```
ssciolla, line 337: Nit-pick: I'm personally not a fan of the mutating version of `drop_duplicates` (with `inplace`). Seems like things are moving toward deprecating this kind of usage: https://github.com/pandas-dev/pandas/issues/16529 

I actually had never used `drop_duplicates` in this way and didn't know about it until I was confused when looking at @jonespm's code on this project: see the closed issue #52. (GitHub issue discovery courtesy of @jonespm).

mivideo/mivideo_extract.py
```diff
@@ -297,31 +296,45 @@ def _makeCourseData(resultDictionaries: Sequence[Dict], categoryFilter: str) ->
...
+            courseData
+                .assign(categories=courseData['categories'].map(lambda c: c.split(',')))
+                .explode('categories')
+        )
 
-        courseData['in_context'] = [
-            c.endswith('>InContext') for c in courseData['course_id']]
+        # remove rows whose category isn't in "Canvas_UMich>site>channels>" hierarchy
+        courseData = courseData[courseData['categories']
+            .map(lambda c: c.startswith('Canvas_UMich>site>channels>'))]
```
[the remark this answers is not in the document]
lsloan: flake8 didn't like it?  I used Pylint, which didn't complain. Codacy seemed OK with it, so I guess you're running flake8 on your own.  I think I don't have a working flake8 installation right now (I'll work on that), so could you tell me what's the error/warning message it gives for this code?

mivideo/mivideo_extract.py
```diff
@@ -297,31 +296,45 @@ def _makeCourseData(resultDictionaries: Sequence[Dict], categoryFilter: str) ->
...
             resultDictionaries, columns=('id', 'categories',)
-        ).rename(columns={'id': 'media_id', 'categories': 'course_id', })
+        ).rename(columns={'id': 'media_id', })
 
-        courseData = courseData.assign(
-            course_id=courseData['course_id'].str.split(',')).explode('course_id')
+        # split category CSVs to multiple new rows
+        courseData = (
+            courseData
+                .assign(categories=courseData['categories'].map(lambda c: c.split(',')))
```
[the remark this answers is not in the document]
lsloan: Same flake8 message as in the other case?

mivideo/mivideo_extract.py
lsloan: Yes, I'm adding the extra delimiter characters to make sure the category string has at least the minimum number of delimiters to split on.  Then we'll end up with an array of at least that many elements.  In the case where we receive some Kaltura category string that's short, like "Language Resource Center" (no subcategory delimiters at all), then this code will return an empty string.  The subsequent code will filter that out.

I'm not sure whether it's a "standard" practice.  I've done this kind of thing before over the years.  It's not necessarily "**_non_**-standard".

mivideo/mivideo_extract.py
lsloan: Yes!  I'm glad you asked.  🙂

`isdecimal()` matches only the characters '0' through '9'.  `isnumeric()` also matches things like super- and subscript digits ('⁰'–'⁹' and '₀'–'₉'; `isdigit()` matches these, too), fractions (e.g., '⅔', '⅖'), and maybe others.  We don't want those characters.

RTFM (no offense) on those functions.  This example summarizes it well: 

```python
def tests(s):
    print('s:', s)
    print('s.isdecimal (): ', s.isdecimal ())
    print('s.isnumeric (): ', s.isnumeric ())
    print('s.isdigit (): ', s.isdigit ())
    print()

tests(u'36246')  #decimal characters
tests(u'⁹')  #superscript digit
tests(u'½¼')  #fractional value
tests(u'③④⑤')  # circled digits
```

mivideo/mivideo_extract.py
lsloan: Ah, too bad.  One of the things I don't like about Pandas is that the code tends to be too wordy with repetitions of the names of the Dataframes and Series.  I wish more functions had the option of using `inplace=`.

With the deprecation looking, I will change this code.

mivideo/mivideo_extract.py
lsloan: FWIW, I don't like logging the warning you suggested because anyone at any time can start adding categories to MiVideo media that doesn't conform to the pattern.  Then the logs will contain a lot of warnings that will worry maintainers, but they turn out to be nothing.

The important thing is that at this point in the code, a media has 𝒏 categories and at least one of them begins with `Canvas_UMich>site>channels>`.  This code is trying to weed out those categories that do not have that.

mivideo/mivideo_extract.py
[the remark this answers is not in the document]
ssciolla: "continuation line under-indented for visual indent flake8(E128)"

mivideo/mivideo_extract.py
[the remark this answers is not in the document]
ssciolla: "continuation line unaligned for hanging indent flake8(E131)"

mivideo/mivideo_extract.py
ssciolla: The code I provided was meant to be an example, not necessarily an exact solution, since I don't know all the ins and outs of this data. You could put it at `DEBUG` level, though that still may be too many log messages, I have no idea.

ssciolla: I'm not going to insist on you making any changes here, as I'm having trouble thinking of cases where this might actually cause problems. I still think something like what I provided is more clear than modifying the input string and assuming cases when an empty string is returned means there weren't `>` delimiters.

mivideo/mivideo_extract.py
ssciolla: Having a hard time imagining these weird values in output, but OK, what you have is fine.

lsloan: Yes, I have a hard time imagining anyone would use those characters in MiVideo media categories, either.  But the point is, they could.  I don't know how well Kaltura prevents those characters.  Should they prevent them?  Maybe not.  Whether or not they're there, we don't want them in this case.  That's why `isdecimal()` is the right thing to use here.

mivideo/mivideo_extract.py
lsloan: Yeah, I know you're not insisting.  I appreciate that.  I also respect your suggestion.  **_Maybe_** my code is a little more like an incantation rather than simple and clear.

If this project weren't ramping down and I had other work to do in this code, then I probably would follow your advice write this in a clearer way.  But now, I don't feel like it's worthwhile to change, test, debug, commit, etc.

mivideo/mivideo_extract.py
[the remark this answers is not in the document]
lsloan: OK.  Pylint had some complaints about my indentations, too.  I corrected them somewhat.  This indentation is mostly done automatically by my IDE (IntelliJ IDEA with Python extension, equiv. to PyCharm), it runs properly, and it seems readable to me, so I'm not worried about this.

It's funny to me that this is an "E" code, which I believe means "error".  It doesn't seem to be an error.  Also, the example they give for E131 isn't like the case I'm using, so it's not clear to me what they want.

I'm going to resolve this and other other flake8 issue.

lsloan:
> The only one that _really_ seems above nit-pick level is about the concatenation of `>>>`, which seems (dare I say) hack-y?

I'm truly honored to have my code called "hack-y".  That's high praise.  😉  I've admired obfuscated C and JAPH examples.  Not that I want to write production code **_that_** weird, but I like that my little example here registered that way.

So, I've agreed with your comment about `inplace=` and replaced that.  I agree in theory that I would make my hack less hack-y in other circumstances.  The flake8 problems don't bother me much, since my IDE and Pylint were fine with those lines.  So, I've resolved most of our conversations.

@ssciolla, would you be willing to approve this PR now?

lsloan:
Yes, basic checks.  The number of records returned by the new code was a little more than the number of records before.  I checked and they were for new media or new placements.

PR #194, opened 2020-06-10
define MiVideo table relationships (#183)

### LINKED ISSUE #183: define MiVideo table relationships
MiVideo tables need foreign key relationships to ensure data integrity.

For example:
* [ ] mivideo_media_created.id → mivideo_media_courses.media_id

---

**_Update_**: This had been discussed by @ssciolla, @lsloan, and others in the past.  It was determined that relationships between MiVideo and other tables would break when updates were made, causing errors.  As Sam notes below, it makes sense to establish the relationship between `mivideo_media_created` and `mivideo_media_courses`.

lsloan:
Resolves #183.

## Summary
Add relationships between `mivideo_media_courses` and `mivideo_media_created` tables.

## Review
I **_humbly_** request that comments on this PR be nested under a single review per reviewer. It helps me when the comments are organized this way, especially if the comments request changes in the code.

If I've requested a review from you but you are definitely unable to review this PR, please leave a comment saying so (please state why you're unable to review, that would be appreciated), tell the other reviewers that the review will be left to them, and remove your name from the list of reviewers.

### DIFF
--- a/db/migrations/0021.add_mivideo_fk.py
+++ b/db/migrations/0021.add_mivideo_fk.py
@@ -15,5 +15,4 @@ steps = [
           ON UPDATE CASCADE
           ON DELETE CASCADE;
     '''),
-
 ]
--- a/db/migrations/0021.add_mivideo_fk.py
+++ b/db/migrations/0021.add_mivideo_fk.py
@@ -1,0 +1,19 @@
+'''
+Migration to add MiVideo table foreign key
+'''
+
+from yoyo import step
+
+__depends__ = {'0016.add_mivideo_creation_tables'}
+
+steps = [
+    step('''
+        ALTER TABLE mivideo_media_courses 
+          ADD CONSTRAINT fk_media_id 
+          FOREIGN KEY (media_id)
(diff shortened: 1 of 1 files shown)


db/migrations/0021.add_mivideo_fk.py
```diff
@@ -0,0 +1,18 @@
...
+steps = [
+    step('''
+        ALTER TABLE mivideo_media_courses 
+          ADD CONSTRAINT fk_media_id 
+          FOREIGN KEY (media_id)
+          REFERENCES mivideo_media_created(id)
+          ON UPDATE CASCADE
+          ON DELETE CASCADE;
+    '''),
+]
```
ssciolla, line 18: Nit-pick: This could just be `step()`, but not a big deal.

ssciolla, line 18: Nit-pick: This could just be `step(...)`, but not a big deal.

ssciolla: Also, if this is a test, I didn't miss your unnecessary trailing comma.

PR #195, opened 2020-06-12
Create a lookup table for LTI placement types (#193)

### LINKED ISSUE #193: Create a lookup table for LTI placement types
It was requested on the DW meeting that we should have a lookup table for LTI tools something like

**lti_type**
```
canvas_id        (int)     # Actual ID from Canvas
des              (varchar) # Description/Name of the tool
rpt_group_nm     (varchar) # Group we define
```

This will map back to the placement id in `lti_placement`.

We'll generate a migration to populate this table. This will replace the environment variable `SUPPORTED_LTI_TOOLS`

jonespm:
Fixes #193

Switches the environment variable for supported placements to be defined
in the database. Populate the new database table and pull the id's out
of it to check against.


lti_placements/canvas_placements.py
```diff
@@ -248,11 +253,11 @@ def main() -> Sequence[DataSourceStatus]:
     This method is invoked when its module is executed as a standalone program.
     '''
 
+    # Get ids for tools in lti_type table as supported tools
```
ssciolla, line 256: Nit-pick: This comment seems a little out of place here. I understand that you're explaining why the `SUPPORTED_LTI_TOOLS` dropped out, but once that's gone, there won't be much context for this. I'd put this wherever you call `get_supported_lti_tools`, or delete it.

db/migrations/0021.add_lti_type.py
```diff
@@ -0,0 +1,23 @@
...
+            rpt_group_nm VARCHAR(200),
+            PRIMARY KEY (canvas_id)
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
+    '''),
+    step('''
+        INSERT INTO lti_type (canvas_id, des, rpt_group_nm)
+        VALUES (25194, 'Zoom', 'Zoom'),
+               (4352, 'Bluejeans', 'Bluejeans')
+    '''),
+]
```
ssciolla, line 23: We should add a FK constraint from `lti_placement` to `lti_type`.

db/migrations/0021.add_lti_type.py
```diff
@@ -0,0 +1,23 @@
...
+            des VARCHAR(200),
+            rpt_group_nm VARCHAR(200),
+            PRIMARY KEY (canvas_id)
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
+    '''),
+    step('''
+        INSERT INTO lti_type (canvas_id, des, rpt_group_nm)
+        VALUES (25194, 'Zoom', 'Zoom'),
+               (4352, 'Bluejeans', 'Bluejeans')
+    '''),
```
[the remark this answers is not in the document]
jonespm: Yeah, we could just manually populate these without the migration but it would probably be nice to keep the migration files up-to-date if we ever do this.

db/db_creator.py
```diff
@@ -101,3 +101,14 @@ def reset_database(self) -> DBCreator:
         '''
         self.drop_records().migrate()
         return self
+
+    def get_pk_values(self, tbl_nm: str, pri_key: str) -> List[Union[int, None]]:
```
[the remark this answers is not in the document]
jonespm: I feel like I just did this to keep the line length down when I was first writing this. I could expand them.

db/migrations/0021.add_lti_type.py
```diff
@@ -0,0 +1,23 @@
...
+from yoyo import step
+
+__depends__ = {'0020.add_lti_usage'}
+
+steps = [
+    step('''
+        CREATE TABLE IF NOT EXISTS lti_type (
+            canvas_id BIGINT NOT NULL,
+            des VARCHAR(200),
+            rpt_group_nm VARCHAR(200),
```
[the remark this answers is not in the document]
jonespm: This was mostly done to match what Jeff has in his data model diagram but without the prefix. 

image

db/migrations/0021.add_lti_type.py
```diff
@@ -0,0 +1,23 @@
...
+            PRIMARY KEY (canvas_id)
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
+    '''),
+    step('''
+        INSERT INTO lti_type (canvas_id, des, rpt_group_nm)
+        VALUES (25194, 'Zoom', 'Zoom'),
```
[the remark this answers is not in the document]
jonespm: I think "Zoom for Health" will be unique enough for that when/if we do add it. 

Zoom == Main Zoom.

db/migrations/0021.add_lti_type.py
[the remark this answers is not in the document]
ssciolla: Okay, if you want to keep it the same to mirror what they have, that's fine with me.

db/migrations/0021.add_lti_type.py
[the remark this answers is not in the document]
ssciolla: Okay.

db/db_creator.py
```diff
@@ -101,3 +101,14 @@ def reset_database(self) -> DBCreator:
...
+        '''
+        Retrieves primary key values from the table. Only works with one primary key.
+        '''
+        pk_values = []
+        conn = self.engine.connect()
+        rs = conn.execute(f"SELECT {pri_key} FROM {tbl_nm}")
```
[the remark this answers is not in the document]
jonespm: Yeah, I looked at these and the problem with sqlalchemy is I didn't see any way you could add in a variable replacement that it didn't quote. 

So like if you want to add in table names or column names this is the only way I saw to do it. Values can be done using a different syntax. 

I guess they figure if you're going to be doing dynamic tables and columns you're doing it correctly.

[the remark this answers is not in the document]
ssciolla: Yep, I've done the similar research about `drop_records`. The main precaution we can take -- without using an ORM -- is validating inputs.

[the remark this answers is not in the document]
jonespm: Yeah, I see what you mean. 

I'm not sure how you'd get the columns for a table. There's probably a query to figure out the primary key for a table that I could call without needing that passed in, this just seemed like a faster fix. 

As you said it's not a huge risk as both of these variables in this context are 100% safe since this these values are part of the code and not something the user can change or inject, but I see the point. 

I think I'll just leave this for now to avoid extra complexity.

lti_placements/canvas_placements.py
jonespm: Yeah that should have been removed

ssciolla:
Looks okay, I will finish reviewing tomorrow morning.

db/migrations/0022.add_lti_type.py
```diff
@@ -0,0 +1,28 @@
...
+            canvas_id INTEGER NOT NULL,
+            des VARCHAR(200),
+            rpt_group_nm VARCHAR(200),
+            PRIMARY KEY (canvas_id)
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
+    '''),
+    step('''
+        ALTER TABLE lti_placement MODIFY placement_type_id INTEGER NULL,
+        ADD FOREIGN KEY (placement_type_id) REFERENCES lti_type(canvas_id) 
+        ON DELETE SET NULL ON UPDATE CASCADE
```
ssciolla, line 21: Question: Not sure I understand the rationale behind `SET NULL` versus `CASCADE`, since if you don't know the `lti_type`, the data becomes much less useful. *Maybe* this leaves the door open to a prior inventory of placements for a tool and then a reassignment of the tool ID once a tool's record is updated with new data. But we're dropping records and repopulating at the moment, so maintaining relationships like that isn't particularly relevant. Regardless, this is an edge case and I don't think it matters particularly, since we're not doing a bunch of one-off database transactions.

lsloan:
This code hasn't yet run in the test environment.  I need to run it there, but this PR needs to be merged.  

@jonespm, can you merge this PR this morning, so I can run it by this afternoon, prior to the planed deployment on Friday morning?

db/migrations/0022.add_lti_type.py
jonespm: Yeah, I didn't know what to do with this. It *probably* doesn't matter as you say right now but I also didn't want someone removing data to cause other data to remove.

I could just leave it as ON DELETE RESTRICT and not allow nulls. It sounds like this is the default so I really could just leave this out and not alter it to null. 

I guess I'll do that.

jonespm: Okay I updated this.

db/migrations/0022.add_lti_type.py
```diff
@@ -0,0 +1,28 @@
...
+            canvas_id INTEGER NOT NULL,
+            des VARCHAR(200),
+            rpt_group_nm VARCHAR(200),
+            PRIMARY KEY (canvas_id)
+        ) ENGINE = InnoDB DEFAULT CHARSET = utf8mb4;
+    '''),
+    step('''
+        ALTER TABLE lti_placement
+        ADD FOREIGN KEY (placement_type_id) REFERENCES lti_type(canvas_id) 
+        ON DELETE RESTRICT ON UPDATE CASCADE
```
lsloan, line 21: According to "MySQL 8.0 Reference Manual  /  ...  /  FOREIGN KEY Constraints" (https://dev.mysql.com/doc/refman/8.0/en/create-table-foreign-keys.html#foreign-key-referential-actions):

> RESTRICT: Rejects the delete or update operation for the parent table. Specifying RESTRICT (or NO ACTION) is the same as omitting the ON DELETE or ON UPDATE clause.

The 5.7 manual says the same.  I think our DBs are 5.7.

jonespm: 'Explicit is better than implicit' right? ;) 

I think I'd still need the clause there for ON UPDATE as I think the default for that is NO ACTION, and I do want it to update if those are updated.

lsloan: Yeah, explicitness usually doesn't hurt.  Unless MySQL's functionality changes later.

I think `on update` and `on delete` can be specified independently.  At least, that's what I gather from the syntax shown in the manual.

Anyway, no big deal.  Resolving…

jonespm: Thanks, it's nice to know the defaults. I didn't think to look those up and these were the values I felt most appropriate. Though realistically neither of these probably matter either way.

PR #199, opened 2020-06-18
Add relax_course_name migration (#198)

### LINKED ISSUE #198: course.name data is not unique
The Canvas data model does not enforce uniqueness on the name property (https://portal.inshosteddata.com/docs#course_dim), but we are doing it in our database. This resulted in a production error when two courses with the same name, but different IDs, were encountered. The uniqueness constraint needs to be removed.

ssciolla:
This PR relaxes the `course.name` field of the `course` table by removing its unique constraint (index) and extending the length to a more flexible `256`. The PR aims to resolve issue #198.

Test: The PR addresses a production integrity error preventing insertion of two courses with the same name in the summer term. Running the `COURSE_INVENTORY` job for that term should confirm that this is working properly.

### DIFF
--- a/db/migrations/0023.relax_course_name.py
+++ b/db/migrations/0023.relax_course_name.py
@@ -1,0 +1,16 @@
+#
+# file: migrations/0023.relax_course_name.py
+#
+from yoyo import step
+
+steps = [
+    step('''
+        ALTER TABLE course
+        MODIFY
+            name VARCHAR(256) NOT NULL;
+    '''),
+    step('''
+        ALTER TABLE course
+        DROP INDEX name;
+    ''')
+]
--- a/db/migrations/0023.relax_course_name.py
+++ b/db/migrations/0023.relax_course_name.py
@@ -6,11 +6,11 @@ from yoyo import step
 steps = [
     step('''
         ALTER TABLE course
-        MODIFY
-            name VARCHAR(256) NOT NULL;
+        DROP INDEX name;
     '''),
     step('''
         ALTER TABLE course
-        DROP INDEX name;
+        MODIFY
+            name VARCHAR(256) NOT NULL;
     ''')
 ]


db/migrations/0023.relax_course_name.py
```diff
@@ -0,0 +1,16 @@
...
+from yoyo import step
+
+steps = [
+    step('''
+        ALTER TABLE course
+        MODIFY
+            name VARCHAR(256) NOT NULL;
+    '''),
+    step('''
+        ALTER TABLE course
```
jonespm, line 13: Do you need to drop the index first before increasing the size?

I believe that on utf8mb4 the max indexable field size is 191 unless you we have the large prefix option.

We could still index first 191 characters though. But maybe an index isn't needed at all for this field really.

ssciolla: Um, I'm dropping the index all together, right? I can re-order them if you think it might be a problem. It wasn't locally using `docker-compose`.

jonespm: Yeah I just didn't know if it would cause an error if you had a value that was too big for the index. I didn't actually test this. If you drop it first it for sure wouldn't be an issue.

It's possible MySQL might just keep the index but keep it at a smaller length until it was dropped.

ssciolla: I re-ordered them. Please take a look.

jonespm (approved):

db/migrations/0023.relax_course_name.py
jonespm: Yup looks fine.

Since these both reference the same table you *could* have just done it in the same `ALTER`. I found out about this in my other PR, I actually wasn't 100% sure. But this is just a nitpick. 

```
    step('''

        ALTER TABLE course
        DROP INDEX name, 
        MODIFY
            name VARCHAR(256) NOT NULL;
    ''')

```

jonespm (approved):

zqian (approved):

